Issue #50, opened 2024-10-02
improve `CONTRIBUTING.md`


jorenham:
- Code style: https://typing.readthedocs.io/en/latest/guides/writing_stubs.html#style-guide
- Stub-writing tips & tricks:
  - `scipy._typing`
  - `numpy.typing` vs [`optype.numpy`](https://github.com/jorenham/optype?tab=readme-ov-file#optypenumpy)
  - how to `@overload` on specific positional-or-keyword parameters
  (see e.g. https://github.com/jorenham/scipy-stubs/pull/153#discussion_r1826597984)

vincent-teh:
Hi there. I was writing Scipy code for my uni project last week, only to realize that the current official Scipy does not have a proper static typing system, leading to my VSCode IntelliJ being unable to work with return value properly. While scrolling through Scipy's issue page, I stumbled across your project and I would like to contribute. However, due to my inexperience, I would like to first ask regarding the type-testing file:

> How to contribute, e.g.
reporting bugs
writing type-tests

How could I run the existing file, since the testing file itself is also a stub file, pyi, and there aren't many resources regarding it. Pardon if the question is quite basic :).

jorenham:
> Pardon if the question is quite basic :).

That's no problem at all; typing in Python is very confusing, and the documentation isn't really beginner friendly.

---

> How could I run the existing file, since the testing file itself is also a stub file, pyi, and there aren't many resources regarding it.

These `.pyi` "stub" files are only read by *static* type-checkers. So at runtime, these are simply ignored. This means that `.pyi` stubs aren't something that can "run" like `.py` scripts.

However, the typical workflow for stubs, isn't all that different from "regular" runtime project. The main difference is that instead of running the unit-tests with e.g. `pytest`, you run `stubtest` (part of `mypy`) instead. That will verify that your `{some_module}.pyi` annotations are compatible with the `{some_module}.py` implementation. 

We also use other tools in `scipy-stubs`, such as `ruff`, `basedpyright` and `basedmypy`, to make sure that the typing is correct. 
See https://github.com/jorenham/scipy-stubs/blob/master/CONTRIBUTING.md#your-first-code-contribution on how you can use `tox` to run all of these locally.

When you submit a PR, all these tools will automatically check you code (with github actions), which can be seen as something like a safety net. But since that takes a while to run, it's easier to run e.g. `poe tox` locally.

[...]

PR #108, opened 2024-10-18
simple contributing guide

jorenham:
this is an initial version; I plan to add more details on typing-specific tips & tricks later

towards #50


pavyamsiri:
This looks good to me. It all makes sense. Keen for the tips and tricks :)

PR #118, opened 2024-10-20
complete `scipy.fft.dct`, and add relevant type aliases

Avasam:
This is the only function used by https://github.com/Toufool/AutoSplit, and I'd gladly replace my local https://github.com/Toufool/AutoSplit/tree/9588834e0881e8c09b449954f4c5c24f0e15bd14/typings/scipy for a community-based one.

I've also added and applied type aliases for "DCT type" and "normalization mode"

I'm not 100% certain about the numpy generic types.

### DIFF
--- a/scipy-stubs/_typing.pyi
+++ b/scipy-stubs/_typing.pyi
@@ -93,6 +93,8 @@ CorrelateMode: TypeAlias = Literal["valid", "same", "full"]
 # scipy literals
 NanPolicy: TypeAlias = Literal["raise", "propagate", "omit"]
 Alternative: TypeAlias = Literal["two-sided", "less", "greater"]
+DCTType: TypeAlias = Literal[1, 2, 3, 4]
+NormalizationMode: TypeAlias = Literal["backward", "ortho", "forward"]
 
 # used in `scipy.linalg.blas` and `scipy.linalg.lapack`
 @type_check_only
--- a/scipy-stubs/fft/_realtransforms.pyi
+++ b/scipy-stubs/fft/_realtransforms.pyi
@@ -1,14 +1,13 @@
 from numpy import float64, generic
 from numpy.typing import NDArray
-
-from scipy._typing import Untyped
+from scipy._typing import DCTType, NormalizationMode, Untyped
 
 def dctn(
     x: Untyped,
-    type: int = 2,
+    type: DCTType = 2,
     s: Untyped | None = None,
     axes: Untyped | None = None,
-    norm: Untyped | None = None,
+    norm: NormalizationMode | None = None,
     overwrite_x: bool = False,
     workers: Untyped | None = None,
     *,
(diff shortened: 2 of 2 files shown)


scipy-stubs/fft/_realtransforms.pyi
```diff
@@ -1,82 +1,84 @@
-from scipy._typing import Untyped
+from numpy import float64, generic
+from numpy.typing import NDArray
```
jorenham, line 2: conventionally `numpy` is imported as module with alias `np`, and `numpy.typing` as `npt`

scipy-stubs/_typing.pyi
```diff
@@ -93,6 +93,8 @@ CorrelateMode: TypeAlias = Literal["valid", "same", "full"]
 # scipy literals
 NanPolicy: TypeAlias = Literal["raise", "propagate", "omit"]
 Alternative: TypeAlias = Literal["two-sided", "less", "greater"]
+DCTType: TypeAlias = Literal[1, 2, 3, 4]
+NormalizationMode: TypeAlias = Literal["backward", "ortho", "forward"]
```
jorenham, line 97: So you know whether these aliases will be useful outside of `scipy.fft`? 
Because if not, then it's a bit cleaner to create a dedicated `scipy.fft._typing` (and add it to [the allowlist](https://github.com/jorenham/scipy-stubs/blob/bc9eba3ad6c2d6c7be69a17dd3e6c8a54a521bf8/tests/stubtest/allowlist.txt#L1-L6))

scipy-stubs/fft/_realtransforms.pyi
```diff
@@ -1,82 +1,84 @@
...
+    norm: NormalizationMode | None = None,
     overwrite_x: bool = False,
     workers: Untyped | None = None,
     orthogonalize: Untyped | None = None,
 ) -> Untyped: ...
 def dct(
-    x: Untyped,
-    type: int = 2,
-    n: Untyped | None = None,
+    x: NDArray[generic],
```
jorenham, line 47: the [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.dct.html) describe `x` as an "`array_like`". So that means that e.g. `[1, 2, 3]` is also allowed here. 

```suggestion
    x: npt.ArrayLike,
```

If you want to, you could narrow this down even further by requiring it to be numeric (i.e. integral, real, or complex), by using the `_ArrayLikeNumber_co` alias from `numpy._typing` instead (don't worry about it being private).

scipy-stubs/fft/_realtransforms.pyi
```diff
@@ -1,82 +1,84 @@
...
     axis: int = -1,
-    norm: Untyped | None = None,
+    norm: NormalizationMode | None = None,
     overwrite_x: bool = False,
-    workers: Untyped | None = None,
-    orthogonalize: Untyped | None = None,
-) -> Untyped: ...
+    workers: int | None = None,
+    orthogonalize: bool | None = None,
+) -> NDArray[float64]: ...
```
jorenham, line 55: If the input is an array of e.g. `float32`, then the output will also be `float32`. 
In the same way, the output could also be complex.
But if the input has a dtype of e.g. `np.int64`, `np.uint8`, or `np.bool_`, then the output will indeed always be of `np.float64`.

I think the cleanest way to go about this is with 2 or 3 overloads, one for integer array-like input, and the other(s) for floating- and complex- array-likes.

jorenham (changes requested):
Thank you for this! Every improvement helps, so it's no problem to specifically target a single function like this 😄 

I left a few suggestions on how to improve this. 
The reason a can't accept this as-is, is because the current `dct` annotations could lead to incorrectly inferred types and false positives. See my inline `dct` comment for why this is.

scipy-stubs/_typing.pyi
jorenham: Oh I just realized; `scipy.fftpack` is likely to benefit from these aliases as well, so this is perfectly fine as-is 👌🏻

scipy-stubs/fft/_realtransforms.pyi
jorenham: this is also why (and only why) the CI is failing btw

scipy-stubs/fft/_realtransforms.pyi
jorenham: As far as I'm concerned, it's also fine if you don't use overloads and `TypeVar`s, and change the return type to `npt.NDArray[np.inexact[Any]]` instead. Even though it's not as "tight" as it could be, it's a lot easier to implement. And all things considered, it'd still be an improvement :)

The underlying thought here is that both `np.floating` and `np.complexfloating` have `np.inexact` as their supertype, so the return type is always correct.

Avasam: I don't mind using overloads, it's just that I'm not sure what's all the numpy dtypes that I need to cover ^^"

> The underlying thought here is that both `np.floating` and `np.complexfloating` have `np.inexact` as their supertype, so the return type is always correct.

There's a catch for return types: This will actually introduce false-positives. For example, I could no longer pass the result of `fft.dct` to `np.median`:

```py
import cv2
import numpy as np
from cv2.typing import MatLike
from scipy import fft

def __cv2_phash(image: MatLike, hash_size: int = 8, highfreq_factor: int = 4):
    img_size = hash_size * highfreq_factor
    image = cv2.cvtColor(image, cv2.COLOR_BGRA2GRAY)
    image = cv2.resize(image, (img_size, img_size), interpolation=cv2.INTER_AREA)
    dct = fft.dct(fft.dct(image, axis=0), axis=1)
    dct_low_frequency = dct[:hash_size, :hash_size]
    median = np.median(dct_low_frequency)
    return dct_low_frequency > median
```
pyright:
```
error: No overloads for "median" match the provided arguments (reportCallIssue)
error: Argument of type "ndarray[Any, dtype[inexact[Incomplete]]]" cannot be assigned to parameter "a" of type "_ArrayLikeFloat_co | _ArrayLikeComplex_co | _ArrayLikeTD64_co | _ArrayLikeObject_co" in function "median"        
    Type "ndarray[Any, dtype[inexact[Incomplete]]]" is not assignable to type "_ArrayLikeFloat_co | _ArrayLikeComplex_co | _ArrayLikeTD64_co | _ArrayLikeObject_co"
      "ndarray[Any, dtype[inexact[Incomplete]]]" is incompatible with protocol "_SupportsArray[dtype[bool_ | integer[Any] | floating[Any]]]"
        "__array__" is an incompatible type
          No overloaded function matches type "() -> ndarray[Any, _DType_co@_SupportsArray]"
      "ndarray[Any, dtype[inexact[Incomplete]]]" is incompatible with protocol "_NestedSequence[_SupportsArray[dtype[bool_ | integer[Any] | floating[Any]]]]"
        "__reversed__" is not present
        "count" is not present
        "index" is not present
    ... (reportArgumentType)
```
mypy:
```
error: Argument 1 to "median" has incompatible type "ndarray[Any, dtype[inexact[Any]]]"; expected "_SupportsArray[dtype[bool_ | integer[Any] | floating[Any]]] | _NestedSequence[_SupportsArray[dtype[bool_ | integer[Any] | floating[Any]]]] | bool | int | float | _NestedSequence[bool | int | float]"  [arg-type]
```

And this isn't even a case that's fixable with an `isinstance` check.

(very similar to this are issues with union return types: https://github.com/python/mypy/issues/1693, both of which could be worked around with https://github.com/python/typing/issues/566 when overloads aren't an option)

jorenham: That's a very good point. I'll definitely keep that in mind when writing/reviewing similar `floating` vs `complexfloating` situations from now on 👌🏻.

That `AnyOf` proposal would indeed be very welcome here. Sometimes you can work around it by using e.g. `T: (str, date)`, but that's limited to within the scope of a function/class/type-alias, and would in this case upcast `datetime` to `date`.
So I really like that idea; thanks for sharing.

jorenham (approved):

scipy-stubs/fft/_realtransforms.pyi
```diff
@@ -1,83 +1,88 @@
...
+    norm: NormalizationMode | None = None,
     overwrite_x: bool = False,
-    workers: Untyped | None = None,
-    orthogonalize: Untyped | None = None,
+    workers: int | None = None,
+    orthogonalize: bool | None = None,
 ) -> Untyped: ...
+
+# We could use overloads based on the type of x to get more accurate return type
+# see https://github.com/jorenham/scipy-stubs/pull/118#discussion_r1807957439
```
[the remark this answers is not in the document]
Avasam: I think it can be done as a follow-up. As you said it's an overall improvement, and it satisfies my needs.

jorenham:
Thanks, @Avasam!

PR #121, opened 2024-10-21
complete `scipy.fftpack`

jorenham:
Every function is now completely typed (no more `Untyped`).
However, there are some oppurtunities for improvement, e.g.:

- input-dependent dtypes of returned arrays
- narrower array-like parameter types, e.g. in case scalar input shoouldn't be allowed


jorenham:
note to self: wait until #118 is merged, then use its `scipy._typing.DCTType` type-alias in `_realtransforms.pyi` (replacing `_TypeKind`)

PR #124, opened 2024-10-21
complete `scipy.signal._savitzky_golay`

pavyamsiri:
Before this PR, the stub file `_savitzky_golay.pyi` had untyped annotations. This PR adds in the missing type annotations and improves previous ones.

The improvements are:
1. Use `op.CanIndex` instead of `int` for the `axis` parameter.
2. Use Literals to type hint certain strings for `use` and `mode` parameters

This file tends to use `int` and `float` a lot which could perhaps be replaced by `AnyInt` and `AnyReal` however, some parameters like `window_length` are used as shapes in calls to functions like `np.zeros` which don't accept `np.bool_`.

For simplicity I left the type annotations as the builtin python types but I am open to using more general type annotations to allow numpy dtypes as well. It just might take a bit more testing to check the allowed types.

### DIFF
--- a/scipy-stubs/signal/_savitzky_golay.pyi
+++ b/scipy-stubs/signal/_savitzky_golay.pyi
@@ -4,7 +4,7 @@ import numpy as np
 import numpy.typing as npt
 import optype as op
 
-_Array_fc_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.inexact[npt.NBitBase]]]
+_Array_f_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.floating[npt.NBitBase]]]
 _Mode: TypeAlias = Literal["mirror", "constant", "nearest", "wrap", "interp"]
 
 def savgol_coeffs(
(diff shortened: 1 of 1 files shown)


scipy-stubs/signal/_savitzky_golay.pyi
```diff
@@ -1,20 +1,27 @@
...
 def savgol_coeffs(
     window_length: int,
     polyorder: int,
     deriv: int = 0,
     delta: float = 1.0,
     pos: int | None = None,
-    use: str = "conv",
-) -> UntypedArray: ...
+    use: Literal["conv", "dot"] = "conv",
+) -> _Array_fc_1d: ...
```
jorenham, line 17: I can't find anything in the docs about complex return types. So I'm missing something, then using `np.floating[Any]` instead here would be best.

pavyamsiri: Thanks for the heads up. I think I just saw that the output was due to some sort of convolution and took the return type from there. I don't think the coefficients of the filter can be complex but I'm also not too well versed on filters.

pavyamsiri: Should be fixed now.

pavyamsiri:
> Unless you want to change something, I think this we can go ahead and merge this :)

Yes should be all done now.

jorenham:
Thanks again @pavyamsiri !

PR #147, opened 2024-10-31
bump to `basedmypy 2.7.0` (`mypy 1.3.0`)

jorenham:
release notes:

- https://github.com/KotlinIsland/basedmypy/releases/tag/v2.7.0
- https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-13

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2,38 +2,43 @@
 
 [[package]]
 name = "basedmypy"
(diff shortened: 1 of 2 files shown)


KotlinIsland:
you should set `basedmypy[faster-cache]`

Issue #99, opened 2024-10-18
🌕 complete `scipy.signal`


jorenham:
it uses `Untyped*` ~662~ ~628~ ~623~ ~517~ ~506~ ~453~ ~296~ ~186~ 156 times

- [x] `_arraytools`
- [x] `_spline_filters` (`_bsplines` in `scipy<1.15`)
- [x] `_czt`
- [x] `_filter_design` - #378
- [x] `_fir_filter_design` - #384
- [x] `_lti_conversion`
- [x] `_ltisys` - #185
- [x] `_max_len_seq_inner`
- [x] `_max_len_seq` - #384
- [x] `_peak_finding` - #87
- [x] `_peak_finding_utils`
- [x] `_savitzky_golay` - #124
- [x] `_short_time_fft` - #226 
- [x] `_signaltools` - #379
- [x] `_spectral`
- [x] `_spectral_py` - #157
- [x] `_spline` - #348
- [x] `_upfirdn`
- [x] `_waveforms`
- [x] `_wavelets` - #195
- [x] `windows._windows` - #153

jorenham:
Once #157 is merged we'll be at 517

jorenham:
@pavyamsiri are you currently working on a particular module or something? Because I'd like to work on this as well, and we should probably avoid a situation where we're both working on the same code.

pavyamsiri:
Not currently no. The remaining modules have a lot of functions in them so I haven't done any work on them yet. I'm not sure when I'll get back to working on this yet so you can have free reign over which ones you want to work on.

jorenham:
> Not currently no. 

Ok, no problem! If there's anything specific you'd like to work on, let me knoow, and I'll "reserve" it for you :)

> The remaining modules have a lot of functions in them so I haven't done any work on them yet. 

Yea there are a couple of real "plus-sized" modules left 😅. It's probably a good idea to split those up. At first glance there appear to be some natural clusters of functions in e.g. `_filter_design` and `_signaltools`, so I guess that could help.

FYI, I'd also gladly accept single-function "coffee-break" PR's 😄.

PR #153, opened 2024-11-02
`signal.windows`: Add type stubs for `_windows.pyi`.

pavyamsiri:
Contributes to completing #99.

This PR adds type stubs for the module `scipy.signal.windows._windows`.

The function overloads are bit dodgy though and I could have maybe used `AnyInt` instead of `int` for some of the arguments.

The reason I didn't use `AnyInt` as of now is because the input arguments `M` go through a private function `_len_guards` which don't allow `np.bool_` as a valid input argument type I believe, as in it just crashes if passed such an arg.

### DIFF
--- a/scipy-stubs/signal/windows/_windows.pyi
+++ b/scipy-stubs/signal/windows/_windows.pyi
@@ -1,4 +1,9 @@
-from scipy._typing import Untyped
+from collections.abc import Sequence
+from typing import Literal, TypeAlias, overload
+from typing_extensions import Unpack
+
+import numpy as np
+from scipy._typing import AnyReal
 
 __all__ = [
     "barthann",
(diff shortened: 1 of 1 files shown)


jorenham:
There are some merge conflicts

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +34,148 @@ __all__ = [
...
+) -> tuple[_Array_f8_2d, _Array_f8_1d]: ...
+@overload
+def dpss(
+    *,
+    M: int,
+    NW: AnyReal,
+    Kmax: None = None,
+    sym: bool = True,
+    norm: Literal[2, "approximate", "subsample "] | None = None,
+    return_ratios: Literal[True] = ...,
```
jorenham, line 175: The story of the `return_ratios` in the previous overload also applies to this one.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +34,148 @@ __all__ = [
...
+    sym: bool = True,
+    norm: Literal[2, "approximate", "subsample "] | None = None,
+    return_ratios: Literal[False] = ...,
+) -> _Array_f8_1d: ...
+@overload
+def dpss(
+    *,
+    M: int,
+    NW: AnyReal,
+    Kmax: int = ...,
```
jorenham, line 162: Unlike `M`, this one looks like it accepts anything with an `__index__` method, i.e. `optype.CanIndex`.

jorenham (commented):
Thanks for this; it's good to see some shape-typing in action 🎉.

I noticed that the `sym` parameters are currently typed as a `bool`, just like the docs say it should be. But I believe that in practice, it could be anything that can be used within an `if` statements, or passed as argument to a `bool`. 
So in similar cases like these, I've recently been using `optype.CanBool` as annotation ([implementation](https://github.com/jorenham/optype/blob/6a153d9c4dde6fefa6bce4a6c679f3e5135cb50b/optype/_core/_can.py#L78-L79)). And while theoretically any `object` should be acceptable (`object` doesn't implement `__bool__`), I thought that annotating them as `_: CanBool` would be a more informative than `_: object`, given the name and all 🤷🏻.
In cases where I had to overload on "truthy" and "falsy" values, I usually went for `Literal[0, False]` and `Literal[1, True]`, as there seems to be a trend within the scipy codebase of using `0` instead of `False`, and `1` instead of `True`.

pavyamsiri:
> There are some merge conflicts

I am actually really confused about this. My point of my commit was to change this file so of course it would conflict, its what I am trying to change!

Must be a git quirk or I just forgot something about how git works.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +34,148 @@ __all__ = [
...
+def lanczos(M: int, *, sym: bool = True) -> _Array_f8_1d: ...
+
+#
+@overload
 def dpss(
-    M,
-    NW,
-    Kmax: Untyped | None = None,
+    *,
+    M: int,
```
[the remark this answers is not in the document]
pavyamsiri: I was considering making a specific int type (private to this module) for this parameter but it seemed like a bit of a pain and `int` seemed like the intention even if it could accept other types.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +34,148 @@ __all__ = [
...
+    NW: AnyReal,
+    Kmax: int = ...,
```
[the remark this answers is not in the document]
pavyamsiri: Honestly I have clue what I am doing with the overloads. When you write an overload for arguments that have default values what values do you put in the overload case where the default value isn't a valid type?

scipy-stubs/signal/windows/_windows.pyi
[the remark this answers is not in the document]
pavyamsiri: Maybe

```python
_WindowLength: TypeAlias = int | np.int16 | np.int32 | np.int64
```
?

jorenham:
The last change made to `_windows.pyi` was two weeks ago, so I suspect you forgot to update your fork before starting this PR then

pavyamsiri:
> The last change made to `_windows.pyi` was two weeks ago, so I suspect you forgot to update your fork before starting this PR then

Okay that makes sense. I think I messed up my branch creation command because I meant to branch off the upstream master using something like `git switch -c my_branch upstream-master`. I thought I had updated my local copy of the upstream master but I guess I didn't.

jorenham:
I often mess it up with git as well; I mostly rely on some vscode plugins for that kinda stuff haha

scipy-stubs/signal/windows/_windows.pyi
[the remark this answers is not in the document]
pavyamsiri: > This is indeed an annoying one to annotate:
> 
> After experimenting a bit with it, it turns out that, besides `builtins.int`, it also accepts signed `short`, `intc`, `intp`, `long` and `longlong`, which boils down to `np.int16 | np.int32 | np.int64`. So unsigned integers aren't included. And for some reason, `np.int8` isn't allowed, so `np.signedinteger[Any]` would be too wide.
> 
> I'm guessing that this is also the case for the `M` param of the other functions in this module

This seems to expose a mistyping in the `linalg` module. The reason why `np.int8` doesn't work is because `dpss` calls `linalg.eigh_tridiagonal` using `M` as part of the `select_range` argument which is currently typed in `scipy-stubs` as

```python
_EigSelectRange: TypeAlias = tuple[spt.AnyReal, spt.AnyReal]
```

which allows `np.int8` even though it seems to crash.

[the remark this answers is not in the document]
pavyamsiri: Looking at the crashing function, it checks that the dtype is one of these `"hilqp"`, which are the types you listed on top, but only if `select="i"`.

pavyamsiri:
I think I addressed everything except not so sure about the overload semantics for the `dpss` function.

scipy-stubs/signal/windows/_windows.pyi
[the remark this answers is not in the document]
jorenham: Good catch! I opened https://github.com/jorenham/scipy-stubs/issues/154 for this.

scipy-stubs/signal/windows/_windows.pyi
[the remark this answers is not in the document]
jorenham: Overloads are kinda similar to pattern matching, but from the perspective of the caller.

So if you have e.g.

```python
def f(x=0, b=False):
    y = (2 << x) - 1
    return x if b else str(x)
```

For simplicity let's only assume builtin types for simplicity, i.e. `x: int` and `b: bool`. 
Because both `x` and `b` can be used positionally or with their keyword, there are four (actually 3, but I'll get to that) possible combinations:

1. `int`: `f()` or `f(x)` or `f(x, False)`
2. `int`: `f(b=False)` or `f(x, b=False)`
3. `str`: `f(b=True)` or `f(x, b=True)`
4. `str`: `f(x, True)`

So with this, the annotations will become

```pyi
from typing import Literal as L, overload

# 1.
@overload
def f(x: int = 0, b: L[False] = False) -> int: ...

# 2. (the `*` indicates that `b` is passed as keyword argument)
@overload
def f(x: int = 0, *, b: L[False]) -> int: ...

# 3.
@overload
def f(x: int = 0, *, b: L[True]) -> str: ... 

# 4. (note that `x` is required in this case, so it has no default)
@overload
def f(x: int, b: L[True]) -> str: ... 
```

Note that the second overload is actually redundant: The first one will already match `f(b=False)` or `f(x, b=False)`, and both overloads have the same return type. So that leaves us with

```pyi
@overload
def f(x: int = 0, b: L[False] = False) -> int: ...
@overload
def f(x: int = 0, *, b: L[True]) -> str: ... 
@overload
def f(x: int, b: L[True]) -> str: ... 
```

In this case, the order doesn't matter, because none of the overloads overlap, so you could equivalently write this as e.g.

```pyi
@overload
def f(x: int, b: L[True]) -> str: ... 
@overload
def f(x: int = 0, *, b: L[True]) -> str: ... 
@overload
def f(x: int = 0, b: L[False] = False) -> int: ...
```

So the order is just a matter of personal preference in this case (although it might slightly affect the performance of type-checker) 🤷🏻.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +34,148 @@ __all__ = [
...
+) -> _Array_f8_1d: ...
+@overload
+def dpss(
+    *,
+    M: int,
+    NW: AnyReal,
+    Kmax: int = ...,
+    sym: bool = True,
+    norm: Literal[2, "approximate", "subsample "] | None = None,
+    return_ratios: Literal[True] = ...,
```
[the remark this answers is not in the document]
jorenham: See https://github.com/jorenham/scipy-stubs/pull/153#discussion_r1826597984

jorenham:
It might be helpful to write a simple type-test for `dpss`, so that you can see the output in each of the main ways that you could call it.
See https://github.com/jorenham/scipy-stubs/tree/master/tests

pavyamsiri:
@jorenham Thanks! The explanation on overloads made a lot of sense. Hopefully the overloads make sense now.

Added some tests for `dpss` as well, like you suggested.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +35,179 @@ __all__ = [
...
+    "ggs",
+    "general hamming",
+    "general_hamming",
+    "kaiser",
+    "ksr",
+    "kaiser bessel derived",
+    "kbd",
+]
+
+def general_cosine(M: _WindowLength, a: Sequence[AnyReal], sym: op.CanBool = True) -> _Array_f8_1d: ...
```
jorenham, line 114: The docs say that `a` should accept array-likes here, but `numpy.ndarray` isn't assignable to `Sequence`. So you could use `numpy._typing._ArrayLikeFloat_co` for this; `Sequence[AnyReal]` and `numpy.ndarray` are both assignable to it.

Alternatively, if you want to enforce `a` to be 1-d, you could keep the `Sequence[AnyReal]`, and union it with `numpy._typing._ArrayLike[np.floating[Any] | np.integer[Any] | np.bool_]`. The latter can be any shape, but at least the `Sequence` part will is enforced as 1d.

jorenham (commented):
apart from one suggestion, this looks about ready to merge

scipy-stubs/signal/windows/_windows.pyi
pavyamsiri: The code I believe conflicts with the docs here because the body is
```python
def general_cosine(M, a, sym=True):
    if _len_guards(M):
        return np.ones(M)
    M, needs_trunc = _extend(M, sym)

    fac = np.linspace(-np.pi, np.pi, M)
    w = np.zeros(M)
    for k in range(len(a)):
        w += a[k] * np.cos(k * fac)

    return _truncate(w, needs_trunc)
```
so I don't think `ArrayLike` works here because no where in the code does it convert the array like into an array. 

I defaulted to `Sequence` because you have to be able to:
- call `len`
- index the array and get some scalar out

The incompatibility with `np.ndarray` was something I forgot about though. This would crash if given a numpy scalar even though that is an array like we would need something like `Sequence[AnyReal]` unioned with the allowed ndarray types.

pavyamsiri: Is there another type alias that is basically a protocol which requires a `__len__` and `__getitem__` implementation? Feel like that is moreso the correct type, although maybe not the intention due to the docs (even if it conflicts with the function implementation).

jorenham: Ah good catch! Then maybe `Sequence[AnyReal] | npt.NDArray[floating[Any]]` 🤔 ?

jorenham: Because the annotations of `numpy.ndarray.__getitem__` currently aren't very good, so I don't think it will work if you'd only use e.g. `optype.CanGetitem` here.

pavyamsiri: I added 1d ndarray types for all the numpy scalar types allowed by `AnyReal` so this should be fixed now.

scipy-stubs/signal/windows/_windows.pyi
```diff
@@ -29,36 +35,185 @@ __all__ = [
...
-def exponential(M: Untyped, center: Untyped | None = None, tau: float = 1.0, sym: bool = True) -> Untyped: ...
-def taylor(M: Untyped, nbar: int = 4, sll: int = 30, norm: bool = True, sym: bool = True) -> Untyped: ...
+_Array_f8_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.float64]]
+_Array_f8_2d: TypeAlias = np.ndarray[tuple[int, int], np.dtype[np.float64]]
+_Weights: TypeAlias = (
+    Sequence[AnyReal]
+    | np.ndarray[tuple[int], np.dtype[np.floating[Any]]]
+    | np.ndarray[tuple[int], np.dtype[np.integer[Any]]]
+    | np.ndarray[tuple[int], np.dtype[np.bool_]]
+)
```
jorenham, line 45: As much as I am a fan of shape-typed arrays, I'm afraid that we can't require specific shapes for input (for now), because numpy has very little support for it.
So for example even simple things like `np.array([1])` currently return `np.ndarray[tuple[int, ...]]` (although I believe that I fixed that, but that'll be in the upcoming 2.2.0 release, and we currently have to support `numpy>=1.24`). 

tldr; for output it's perfectly fine to use shape-typing, but for input it's too restrictive at the moment.

pavyamsiri: Okay got it I will change it.

jorenham (approved):

jorenham:
thanks @pavyamsiri !

scipy-stubs/signal/windows/_windows.pyi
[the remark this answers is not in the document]
jorenham: Fixed in https://github.com/jorenham/scipy-stubs/pull/165

PR #157, opened 2024-11-04
`signal`: Add type stubs to `_spectral_py.pyi`.

pavyamsiri:
Contributes to completing #99.

I added type stubs to the stub file `signal/_spectral_py.pyi` and tests for the function overloads I added.

## Caveats

I discovered that the function `lombscargle` was recently updated to include new parameters which also expands the possible return types (see PR scipy/scipy#21277). It also adds inline type stubs which might be an issue.

I did not type stub against this new version though because I think it is still in the dev branch and is not part of a release.

Also I saw that the function `periodogram`'s docs don't specify that `None` is allowed for the argument `window` but the body explicitly checks for `None`. Therefore I allowed `None` to be passed in as well. This is a bit annoying because the type annotation is no longer consistent with other `window` arguments in other functions.

I could have perhaps used more shaped typing but its a bit hard to verify because the functions are a bit complex, so I just left things as any shape.

### DIFF
--- a/scipy-stubs/signal/_spectral_py.pyi
+++ b/scipy-stubs/signal/_spectral_py.pyi
@@ -1,5 +1,5 @@
 from collections.abc import Callable
-from typing import Any, Literal, TypeAlias
+from typing import Any, Literal, TypeAlias, overload
 from typing_extensions import Unpack
 
 import numpy as np
@@ -11,7 +11,10 @@ from scipy.signal.windows._windows import _Window, _WindowNeedsParams
 
 __all__ = ["check_COLA", "check_NOLA", "coherence", "csd", "istft", "lombscargle", "periodogram", "spectrogram", "stft", "welch"]
 
+_Array_f8: TypeAlias = npt.NDArray[np.float64]
 _Array_f8_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.float64]]
+_ArrayReal: TypeAlias = npt.NDArray[np.floating[Any]]
+_ArrayComplex: TypeAlias = npt.NDArray[np.complexfloating[Any, Any]]
 
 _GetWindowArgument: TypeAlias = _Window | tuple[_Window | _WindowNeedsParams, Unpack[tuple[object, ...]]]
 _Detrend: TypeAlias = Literal["literal", "constant", False] | Callable[[npt.NDArray[np.generic]], npt.NDArray[np.generic]]
@@ -37,7 +40,7 @@ def periodogram(
     return_onesided: op.CanBool = True,
     scaling: _Scaling = "density",
     axis: op.CanIndex = -1,
-) -> tuple[npt.NDArray[np.float64], npt.NDArray[np.floating[Any]]]: ...
+) -> tuple[_Array_f8, _ArrayReal]: ...
 def welch(
     x: _ArrayLikeComplex_co,
     fs: AnyReal = 1.0,
@@ -50,7 +53,7 @@ def welch(
     scaling: _Scaling = "density",
     axis: op.CanIndex = -1,
     average: _Average = "mean",
-) -> tuple[npt.NDArray[np.float64], npt.NDArray[np.floating[Any]]]: ...
+) -> tuple[_Array_f8, _ArrayReal]: ...
 def csd(
     x: _ArrayLikeComplex_co,
     y: _ArrayLikeComplex_co,
@@ -64,20 +67,59 @@ def csd(
     scaling: _Scaling = "density",
     axis: op.CanIndex = -1,
     average: _Average = "mean",
-) -> tuple[npt.NDArray[np.float64], npt.NDArray[np.complexfloating[Any, Any]]]: ...
+) -> tuple[_Array_f8, _ArrayComplex]: ...
+
+#
+@overload
 def spectrogram(
-    x: Untyped,
-    fs: float = 1.0,
-    window: Untyped = ("tukey", 0.25),
-    nperseg: int | None = None,
-    noverlap: int | None = None,
-    nfft: int | None = None,
-    detrend: str | Literal[False] | UntypedCallable = "constant",
-    return_onesided: bool = True,
-    scaling: str = "density",
-    axis: int = -1,
-    mode: str = "psd",
-) -> Untyped: ...
+    x: npt.NDArray[np.generic],
+    fs: AnyReal = 1.0,
+    window: _GetWindowArgument | _ArrayLikeFloat_co = ("tukey", 0.25),
+    nperseg: AnyInt | None = None,
+    noverlap: AnyInt | None = None,
+    nfft: AnyInt | None = None,
+    detrend: _Detrend = "constant",
+    return_onesided: op.CanBool = True,
+    scaling: _Scaling = "density",
+    axis: op.CanIndex = -1,
+    mode: Literal["psd", "magnitude", "angle", "phase"] = "psd",
+) -> tuple[_Array_f8, _Array_f8, _ArrayReal]: ...
+
+# complex mode (positional)
+@overload
+def spectrogram(
+    x: npt.NDArray[np.generic],
+    fs: AnyReal,
+    window: _GetWindowArgument | _ArrayLikeFloat_co,
+    nperseg: AnyInt | None,
+    noverlap: AnyInt | None,
+    nfft: AnyInt | None,
+    detrend: _Detrend,
+    return_onesided: op.CanBool,
+    scaling: _Scaling,
+    axis: op.CanIndex,
+    mode: Literal["complex"],
+) -> tuple[_Array_f8, _Array_f8, _ArrayComplex]: ...
+
+# complex mode (keyword)
+
+@overload
+def spectrogram(
+    x: npt.NDArray[np.generic],
+    fs: AnyReal = 1.0,
+    window: _GetWindowArgument | _ArrayLikeFloat_co = ("tukey", 0.25),
+    nperseg: AnyInt | None = None,
+    noverlap: AnyInt | None = None,
+    nfft: AnyInt | None = None,
+    detrend: _Detrend = "constant",
+    return_onesided: op.CanBool = True,
+    scaling: _Scaling = "density",
+    axis: op.CanIndex = -1,
+    *,
+    mode: Literal["complex"],
+) -> tuple[_Array_f8, _Array_f8, _ArrayComplex]: ...
+
+#
 def check_COLA(
     window: _GetWindowArgument | _ArrayLikeFloat_co,
     nperseg: AnyInt,
--- a/scipy-stubs/signal/_spectral_py.pyi
+++ b/scipy-stubs/signal/_spectral_py.pyi
@@ -17,6 +17,7 @@ _ArrayFloat: TypeAlias = npt.NDArray[np.float32 | np.float64 | np.longdouble]
 _ArrayComplex: TypeAlias = npt.NDArray[np.complex64 | np.complex128 | np.clongdouble]
 
 _GetWindowArgument: TypeAlias = _Window | tuple[_Window | _WindowNeedsParams, Unpack[tuple[object, ...]]]
+_WindowLike: TypeAlias = _GetWindowArgument | _ArrayLikeFloat_co
 _Detrend: TypeAlias = Literal["literal", "constant", False] | Callable[[npt.NDArray[np.generic]], npt.NDArray[np.generic]]
 _Scaling: TypeAlias = Literal["density", "spectrum"]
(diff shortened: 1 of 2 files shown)


jorenham:
> I did not type stub against this new version though because I think it is still in the dev branch and is not part of a release.

Yea, that'll be in the `scipy 1.15.0` release (see https://discuss.scientific-python.org/t/proposed-release-schedule-for-scipy-1-15-0/1502 for the release schedule).

---

> Also I saw that the function `periodogram`'s docs don't specify that `None` is allowed for the argument `window` but the body explicitly checks for `None`. Therefore I allowed `None` to be passed in as well. This is a bit annoying because the type annotation is no longer consistent with other `window` arguments in other functions.

Yea good catch. This is not the first time that the docs are incorrect unfortunately. And I agree with you judgement call here, because as stated in our [code of conduct](https://github.com/jorenham/scipy-stubs/blob/master/CODE_OF_CONDUCT.md):

> Practicality beats purity
> Consistency beats practicality
> Validity beats consistency

---

> I could have perhaps used more shaped typing but its a bit hard to verify because the functions are a bit complex, so I just left things as any shape.

Yea it's usually better to avoid guessing the return type, and go with the "safe" option. Plus, shape-typing is somewhere at the bottom of the (hypothetical) priority list, so no worries.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,209 @@
...
+    nperseg: AnyInt,
+    noverlap: AnyInt,
+    tol: AnyReal = 1e-10,
+) -> bool: ...
+def check_NOLA(
+    window: _GetWindowArgument | _ArrayLikeFloat_co,
+    nperseg: AnyInt,
+    noverlap: AnyInt,
+    tol: AnyReal = 1e-10,
+) -> bool: ...
```
jorenham, line 133: I quickly tried these, and they returned `np.bool`. But I haven't checked whether they could also return `builtins.bool` in some cases.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,209 @@
-from typing import Literal
+from collections.abc import Callable
+from typing import Any, Literal, TypeAlias, overload
+from typing_extensions import Unpack
 
-from scipy._typing import Untyped, UntypedCallable
+import numpy as np
+import numpy.typing as npt
+import optype as op
+from numpy._typing import _ArrayLikeComplex_co, _ArrayLikeFloat_co
```
jorenham, line 8: It's a bit awkward, but `_ArrayLikeComplex_co` is equivalent to `_ArrayLikeNumber_co`, since `np.integer[Any] | np.floating[Any] | np.complexfloating[Any, Any]` is equivalent to `np.number[Any]`. 

So I usually prefer the more concise `_ArrayLikeNumber_co`, because that could help with type-checker performance, and makes the potential error messages of type-checkers ~more readable~ less unreadable.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,209 @@
...
+# input_onesided is `True`
+@overload
+def istft(
+    Zxx: _ArrayLikeComplex_co,
+    fs: AnyReal = 1.0,
+    window: _GetWindowArgument | _ArrayLikeFloat_co = "hann",
+    nperseg: AnyInt | None = None,
+    noverlap: AnyInt | None = None,
+    nfft: AnyInt | None = None,
+    input_onesided: Literal[True] = True,
```
jorenham, line 158: I've seen quite a bit of internal scipy code that uses `1` for `True` and `0` for `False` (probably written C-developer 😋). 
So in case you want to allow that, you could include a `1` in this literal (and a `0` in the `False` literals below). But as far as I'm concerned, it's also fine to leave it like this, and just require users to do as the docs say they should 🤷🏻.

tests/signal/test_spectral.pyi
```diff
@@ -0,0 +1,65 @@
...
+import numpy as np
+import numpy.typing as npt
+from scipy.signal import istft, spectrogram
+
+_Array_f8: TypeAlias = npt.NDArray[np.float64]
+_ArrayReal: TypeAlias = npt.NDArray[np.floating[Any]]
+_ArrayComplex: TypeAlias = npt.NDArray[np.complexfloating[Any, Any]]
+
+# test spectrogram function overloads
+assert_type(spectrogram(np.linspace(200, 300, 256)), tuple[_Array_f8, _Array_f8, _ArrayReal])
```
jorenham, line 13: Instead of these repeated `linspace`s, you could also exploit the fact that we're in a `.pyi`, and use a "fake" variable, e.g.

```pyi
import optype.numpy as onpt

array_1d_f8: onpt.Array[tuple[Literal[256]], np.float64]
```

or similarly, you could go cover a whole range of inputs with e.g.

```pyi
arraylike_1d_f: list[float] | onpt.CanArray[tuple[int], np.dtype[np.floating[Any]]]
```

(but that's probably not needed here)

tests/signal/test_spectral.pyi
```diff
@@ -0,0 +1,65 @@
...
+assert_type(spectrogram(np.linspace(200, 300, 256), mode="psd"), tuple[_Array_f8, _Array_f8, _ArrayReal])
+assert_type(spectrogram(np.linspace(200, 300, 256), mode="magnitude"), tuple[_Array_f8, _Array_f8, _ArrayReal])
+assert_type(spectrogram(np.linspace(200, 300, 256), mode="angle"), tuple[_Array_f8, _Array_f8, _ArrayReal])
+assert_type(spectrogram(np.linspace(200, 300, 256), mode="phase"), tuple[_Array_f8, _Array_f8, _ArrayReal])
```
jorenham, line 17: You could test these different `mode` values in a single call by extracting the values as a "gradual type" like

```
spectrogram_mode_real: Literal["psd", "magnitude", ...]
```

So that's the main advantage of using `.pyi` for type-testing, in case you were wondering :)

jorenham (commented):
Nice job on the type-tests! I should probably do that more often as well haha.

I left a couple of small suggestions, but in general this looks very good to me 👌🏻.

scipy-stubs/signal/_spectral_py.pyi
pavyamsiri: I think they can only return `np.bool` like you said because the only return statement is from a numpy comparison.

pavyamsiri:
I think all the issues should be addressed now.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,209 @@
...
+    nfft: AnyInt | None = None,
+    detrend: _Detrend = "constant",
+    return_onesided: op.CanBool = True,
+    scaling: _Scaling = "density",
+    axis: op.CanIndex = -1,
+    mode: Literal["psd", "magnitude", "angle", "phase"] = "psd",
+) -> tuple[_Array_f8, _Array_f8, _ArrayReal]: ...
+
+# complex mode (positional)
+@overload
```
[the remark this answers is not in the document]
jorenham: I was thinking of placing it like `@overload  # ...`, but this apparently also works 👍🏻

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,206 @@
...
+    mode: Literal["complex"],
+) -> tuple[_Array_f8, _Array_f8, _ArrayComplex]: ...
+
+#
+def check_COLA(
+    window: _GetWindowArgument | _ArrayLikeFloat_co,
+    nperseg: AnyInt,
+    noverlap: AnyInt,
+    tol: AnyReal = 1e-10,
+) -> np.bool: ...
```
jorenham, line 125: I just now noticed, but `numpy.bool` was added in NumPy `2.0`, and we still have to support `>=1.24`, so we have to use `np.bool_` until we can drop support for `numpy<2`.

```suggestion
) -> np.bool_: ...
```

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,206 @@
...
+def check_NOLA(
+    window: _GetWindowArgument | _ArrayLikeFloat_co,
+    nperseg: AnyInt,
+    noverlap: AnyInt,
+    tol: AnyReal = 1e-10,
+) -> np.bool: ...
```
jorenham, line 131:
```suggestion
) -> np.bool_: ...
```

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,206 @@
...
+from scipy._typing import AnyInt, AnyReal
+from .windows._windows import _Window, _WindowNeedsParams
 
 __all__ = ["check_COLA", "check_NOLA", "coherence", "csd", "istft", "lombscargle", "periodogram", "spectrogram", "stft", "welch"]
 
-def lombscargle(x: Untyped, y: Untyped, freqs: Untyped, precenter: bool = False, normalize: bool = False) -> Untyped: ...
+_Array_f8: TypeAlias = npt.NDArray[np.float64]
+_Array_f8_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.float64]]
+_ArrayFloat: TypeAlias = npt.NDArray[np.float32 | np.float64 | np.float128]
+_ArrayComplex: TypeAlias = npt.NDArray[np.complex64 | np.complex128 | np.complex256]
```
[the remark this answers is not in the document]
pavyamsiri: The way I determined these dtypes was I iterated over all the dtypes available to me on my system and then casted the input arrays as those dtypes. Collected all the output dtypes into a set and then checked to see what was outputted. I guess numpy prefers using numbered dtypes over the more ambiguous ones.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,206 @@
...
-    window: str = "boxcar",
-    nfft: int | None = None,
-    detrend: str | Literal[False] | UntypedCallable = "constant",
-    return_onesided: bool = True,
-    scaling: str = "density",
-    axis: int = -1,
-) -> Untyped: ...
+    x: _ArrayLikeNumber_co,
+    fs: AnyReal = 1.0,
+    window: _GetWindowArgument | _ArrayLikeFloat_co | None = "boxcar",
```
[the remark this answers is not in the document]
pavyamsiri: That's a good suggestion.

scipy-stubs/signal/_spectral_py.pyi
```diff
@@ -1,97 +1,206 @@
...
-def lombscargle(x: Untyped, y: Untyped, freqs: Untyped, precenter: bool = False, normalize: bool = False) -> Untyped: ...
+_Array_f8: TypeAlias = npt.NDArray[np.float64]
+_Array_f8_1d: TypeAlias = np.ndarray[tuple[int], np.dtype[np.float64]]
+_ArrayFloat: TypeAlias = npt.NDArray[np.float32 | np.float64 | np.float128]
+_ArrayComplex: TypeAlias = npt.NDArray[np.complex64 | np.complex128 | np.complex256]
+
+_GetWindowArgument: TypeAlias = _Window | tuple[_Window | _WindowNeedsParams, Unpack[tuple[object, ...]]]
+_Detrend: TypeAlias = Literal["literal", "constant", False] | Callable[[npt.NDArray[np.generic]], npt.NDArray[np.generic]]
+_Scaling: TypeAlias = Literal["density", "spectrum"]
+_LegacyScaling: TypeAlias = Literal["psd", "spectrum"]
```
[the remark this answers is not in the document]
pavyamsiri: This is because of the documentation for `stft` and `istft`. See https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.stft.html

The `scaling` parameter type differs from the non-legacy functions so I decided to call this type `_LegacyScaling` to reflect that this is scaling argument from a legacy function.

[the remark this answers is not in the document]
pavyamsiri: > This is because of the documentation for `stft` and `istft` says that these functions are legacy. See https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.stft.html
> 
> The `scaling` parameter type differs from the non-legacy functions so I decided to call this type `_LegacyScaling` to reflect that this is scaling argument from a legacy function.

pavyamsiri:
Good now?

jorenham (approved):

jorenham:
Thanks again, @pavyamsiri 😄

jorenham:
@pavyamsiri Do you think that the "phase" of `scipy.signal` in the `README.md` could be updated from 🌒 to 🌓?

PR #174, opened 2024-11-16
use `uv` as package manager

### LINKED ISSUE #49: migrate from `poetry` to `uv`
[poe doesn't require `poetry`](https://poethepoet.natn.io/guides/without_poetry.html), so there's no need to migrate the current `poe` commands ([but we probably will](https://github.com/jorenham/scipy-stubs/issues/216)).

miloth:
Hi @jorenham , this aims at solving #49. The main changes are:
- Package manager moved to `uv`.
- ~Command runner changed to [`just`](https://github.com/casey/just)~.
- `tox` and `pre-commit` use `uv` as venv backend.

~It is a proof of concept, untested on GHA, hence the "draft" status. I just wanted to check if changing the command runner was an acceptable trade-off before continuing the work.~

Resolves #49


jorenham:
Thanks a lot for this! 
I wanted to see what the CI thought of it, so removed the draft status for a bit.

jorenham:
What are the advantages for using `just` over `poe`? 
FWIW; I believe that `poe` (unlike the name suggests) doesn't require `poetry` to work

miloth:
Yeah, I see the problem. I think it's related to how `uv` installs the packages in the venv. I came up with a workaround in the latest commit, basically forcing the install as a built wheel. Now the `stubtest` CI passes.

Going back to the choice between `poe` and `just`. I believe using Python as a command runner is a bit unnecessary. You are adding the complexity of an additional layer of interpreter for no added benefit. I prefer `just`, as it is a compiled tool, so much more self-contained. In the end though, the commands specified here are quite simple, so the choice will not have a massive impact. In the end, the `uv` team are considering their own command runner, so it could all end up under one tool anyways.

jorenham:
Ok that makes sense 👌🏻. I'm not familiar with `just`, but judging from its popularity, it's pretty solid.

BTW, don't forget update the `CONTRIBUTING.md`, which explicitly mentions `poe` now.

scipy-stubs/py.typed
jorenham, line 1: from https://typing.readthedocs.io/en/latest/spec/distributing.html#type-information-in-libraries:

> For stub-only packages adding a `py.typed` marker is not needed since the name `*-stubs` is enough to indicate it is a source of typing information.

miloth:
The contribution docs should be sorted now. The things I think are worth considering in this PR are:
- How the command runner and `tox` work together. The commands here are quite simple, so I think it could be feasible to get rid of them and put all the config in `tox`. I'll create a child PR into this one to showcase how. Leveraging the `tox-uv` plugin, it should work fairly nicely the CIs.
- I'd like to investigate a bit better why the stubs are not recognized when installed with `uv sync`...
- How do you feel about dividing the `pyproject.toml` in smaller self-contained sections? It's 300+ lines, so it could be a bit difficult finding specific pieces. I was thinking about taking out `ruff.toml` and `tox.toml`. I can do this in a separate PR, to have one per topic.

scipy-stubs/py.typed
miloth: Always learning new things 😆

jorenham:
> * How do you feel about dividing the `pyproject.toml` in smaller self-contained sections? It's 300+ lines, so it could be a bit difficult finding specific pieces. I was thinking about taking out `ruff.toml` and `tox.toml`. I can do this in a separate PR, to have one per topic.

Hmm, I see what you mean, but I usually try to minimize the amount of (config) files in the repo root. Having a central place for configuration can also be helpful in my experience, which can be especially practical when we change a config option that affects multiple tools (e.g. mypy and pyright).

Perhaps it could help if we'd add a comment between sections, so that they're easier to visually distinguish 🤔

jorenham:
Hmm, I see that `just` doesn't automatically run in the `uv` environment. So because`uv run just stubtest` kinda defeats the purpose; maybe we could prefix the `just` commands with a `uv run` in the `justfile`? That should also help simplify the `ci.yml` workflow.

Oh and speaking of GHA, the `just` docs mention the `extractions/setup-just@v2` and `taiki-e/install-action@just` actions. So maybe one of those could help as well.

---

In the meantime I upgraded ruff and fixed the new errors, which is what's causing the merge conflicts, as this PR also included some fixes for those ruff errors. Anyway; there's no need to worry about those ruff errors now; and you can just reset those `.pyi` files to get rid of the conflicts.

---


> * How the command runner and `tox` work together. The commands here are quite simple, so I think it could be feasible to get rid of them and put all the config in `tox`. I'll create a child PR into this one to showcase how. Leveraging the `tox-uv` plugin, it should work fairly nicely the CIs.

That sounds interesting; looking forward to it :)

---

> * I'd like to investigate a bit better why the stubs are not recognized when installed with `uv sync`...

My gut feeling tells me it has something to do with hatch (and I based that on totally nothing 👌🏻). So perhaps switching to another build tool could help with this?

lucascolley:
> `uv run just stubtest`

just to throw another hat into the ring 😅 with https://pixi.sh this would look like `pixi run stubtest` or `pixi r stubtest`.

Probably good to migrate to `uv` first though, then only consider `pixi` if there would be further benefits. I assume some of the dependencies here are PyPI-only anyway, so maybe the benefits of access to the Conda ecosystem won't be substantial.

jorenham:
There's also [`spin`](https://github.com/scientific-python/spin) of course, which is what `numpy` and `scipy` use

lucascolley:
`scipy` _almost_ uses spin 😅 https://github.com/scipy/scipy/pull/21674 (and that is for just tasks rather than dependency management as well)

jorenham:
> The contribution docs should be sorted now. The things I think are worth considering in this PR are:
> 
> * How the command runner and `tox` work together. The commands here are quite simple, so I think it could be feasible to get rid of them and put all the config in `tox`. I'll create a child PR into this one to showcase how. Leveraging the `tox-uv` plugin, it should work fairly nicely the CIs.
> * I'd like to investigate a bit better why the stubs are not recognized when installed with `uv sync`...
> * How do you feel about dividing the `pyproject.toml` in smaller self-contained sections? It's 300+ lines, so it could be a bit difficult finding specific pieces. I was thinking about taking out `ruff.toml` and `tox.toml`. I can do this in a separate PR, to have one per topic.

@miloth I also don't mind merging the UV part of this PR as it currently is. We could then reconsider replacing `poe` in a follow-up PR.

miloth:
Hi @jorenham , I've just come back from a long work trip, no real time to do some fun coding. I'll clean it up next week if you want. Or feel free to do as you suggested in the meantime!

jorenham:
> I've just come back from a long work trip, no real time to do some fun coding. I'll clean it up next week if you want.

No worries; I'm just happy to see that you're still interested :)

In the meantime some merge conflicts have accumulated, but those shouldn't be difficult to deal with.

> Or feel free to do as you suggested in the meantime!

What suggestion of mine are you referring to here?

jorenham:
Thanks @miloth!

PR #226, opened 2024-11-29
`signal`: Replace the final `Untyped` in `_short_time_fft.pyi`.

pavyamsiri:
This contributes to closing #99.

I saw that there was a single `Untyped` left so here is a PR to fix that.


jorenham (approved):

jorenham:
Thanks Pavadol :)

PR #249, opened 2024-12-02
Add a highlights section in `README.md`

### LINKED ISSUE #248: Document the shape-typing capabilities of `scipy-stubs`
Many (if not most) of the `scipy-stubs` function signatures that accept or return array-likes are *shape-typed*. These either enforce the input to be of a certain shape (e.g. 2-D input accepts `list[list[float]]` but rejects `list[float]`), or return arrays of a specific shape (e.g. a `float64` matrix would be `onp.Array2D[np.float64]` i.e. `ndarray[tuple[int, int], np.dtype[np.float64]]`).

To my knowledge, this is the first application of numpy's shape-typing on such a scale, and is even better at shape-typing than `numpy` itself (thanks to [`optype`](https://github.com/jorenham/optype)).

jorenham:
It's a bit flashy, but all very true.

closes #248

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <h1 align="center">scipy-stubs</h1>
 
 <p align="center">
-    Typing stubs for <a href="https://github.com/scipy/scipy">SciPy</a>.
+    Precise type hints for <i>all</i> of <a href="https://github.com/scipy/scipy">SciPy</a>.
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


jorenham:
In case anyone has some thoughts on this, please let me know (as I'm not the best writer). I'll leave it open for a one or two days.

jorenham:
@lucascolley, any thoughts on this?

README.md
```diff
@@ -65,6 +65,32 @@
 
 ______________________________________________________________________
 
+## Highlights
+
+- Works out-of-the-box
+  - all that's needed is to [install `scipy-stubs`](#installation)
+  - does not require a `mypy` plugin or other configuration
+  - available on [PyPI](https://pypi.org/project/scipy-stubs/) and [Anaconda](https://anaconda.org/conda-forge/scipy-stubs)
```
lucascolley, line 73: It is more accurate to say conda-forge here - you can access conda-forge packages independently from anaconda.org, for example: https://prefix.dev/channels/conda-forge/packages/scipy-stubs. Probably fine to still link to the anaconda.org page but either would do.

```suggestion
  - available on [PyPI](https://pypi.org/project/scipy-stubs/) and [conda-forge](https://anaconda.org/conda-forge/scipy-stubs)
```

jorenham:
Thanks @lucascolley :)

PR #259, opened 2024-12-04
dynamic minimum requirements

miloth:
Small scripts to automatically fetch the minimum versions for Python and `numpy`. It adds a separate CI that runs types and tests using `basedpyright` and `pytest` respectively.


jorenham:
Why `pytest` 🤔 ?

tools/__init__.py
jorenham, line 1: this shouldn't be needed

pyproject.toml
```diff
@@ -52,10 +52,19 @@ typecheck = [
 codegen = [
     "libcst>=1.5.1,<2",
 ]
+tools = [
+    "requests~=2.0",
+    "tomli>=2.2.1 ; python_full_version < '3.12'",
+]
```
jorenham, line 58: Since it's a standalone script, [PEP 722](https://peps.python.org/pep-0722/) might be more appropriate. It's even supported by `uv` (https://github.com/astral-sh/uv/issues/7396) by running `uv run --script tools/<long name>.py`

tools/fetch_minimum_requirements.py
jorenham, line 1: `tools` as directory name is a bit vague, and might even be confused with a bundled package or something. I've seen other stub packages such as [`pandas-stubs`](https://github.com/pandas-dev/pandas-stubs) and [`typeshed`](https://github.com/python/typeshed) use `scripts` for this purpose (which is still kinda vague imo, but it's at least convential).

jorenham: As for the script name itself, how about something a more concise, such as `version_bounds.py` or something? It shouldn't matter too much as it's only used internally, but it could help with readability and avoid typo's.

miloth:
> Why pytest 🤔 ?

To check on the methods written to automate the code gen and the CIs and make sure they don't break in the future.

jorenham (commented):
With [`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) you could simplify this quite a bit:

```pycon
>>> import importlib.metadata as impmeta
>>> impmeta.metadata("scipy")["Requires-Python"]
'>=3.10'
>>> [req for req in meta.requires("scipy") if req.startswith("numpy") and not " extra " in req][0]
'numpy<2.3,>=1.23.5'
```

jorenham:
> > Why pytest 🤔 ?
> 
> To check on the methods written to automate the code gen and the CIs and make sure they don't break in the future.

Sure, the codegen scripts could probably do with some testing. But I suppose that something simple like `doctest` could also suffice for that. At this point I don't see any need for having `pytest` in the pipeline. So I think that it's best to keep pytest out of the infrastructure. If we at some point in the future decide we want to use it after all, then we can just re-introduce pytest at that point.

jorenham:
> > > Why pytest 🤔 ?
> > 
> > 
> > To check on the methods written to automate the code gen and the CIs and make sure they don't break in the future.
> 
> Sure, the codegen scripts could probably do with some testing. But I suppose that something simple like `doctest` could also suffice for that. At this point I don't see any need for having `pytest` in the pipeline. So I think that it's best to keep pytest out of the infrastructure. If we at some point in the future decide we want to use it after all, then we can just re-introduce pytest at that point.

... or were you actually planning on writing unit tests for the codemods yourself?

miloth:
> > > > Why pytest 🤔 ?
> > > 
> > > 
> > > To check on the methods written to automate the code gen and the CIs and make sure they don't break in the future.
> > 
> > 
> > Sure, the codegen scripts could probably do with some testing. But I suppose that something simple like `doctest` could also suffice for that. At this point I don't see any need for having `pytest` in the pipeline. So I think that it's best to keep pytest out of the infrastructure. If we at some point in the future decide we want to use it after all, then we can just re-introduce pytest at that point.
> 
> ... or were you actually planning on writing unit tests for the codemods yourself?

Not in this PR. The only (very simple) tests added here are for the functions that fetch the minimum versions.

pyproject.toml
miloth: I ended up moving to `importlib`, as you suggested above. That makes the whole script rely on standard libraries, freeing up the script requirement. Also, it needs to be run in the main env, as `importlib` needs to fetch the currently installed `scipy`.

scripts/version_bounds.py
```diff
@@ -0,0 +1,18 @@
...
+    if "<" in raw_version:
+        raise NotImplementedError("Version specifier with upper bound not yet supported!")
+    return raw_version.replace(">=", "").replace("~=", "")
+
+
+def scipy_minimum_numpy() -> str:
+    """Fetch the minimum numpy version required by the current scipy package."""
+    np_minimum_dep = next(req for req in importlib.metadata.requires("scipy") if req.startswith("numpy") and " extra " not in req)
+    np_minimum_dep = next(ver for ver in np_minimum_dep.split(",") if ">" in ver)
+    return np_minimum_dep.replace("numpy", "").replace(">=", "").replace(">", "")
```
jorenham, line 18: how about something like

```py
def main() -> None:
    import sys

    match sys.argv:
        case [_, "python"]:
            print(scipy_minimum_python())
        case [_, "numpy"]:
            print(scipy_minimum_numpy())
        case _ as badargs:
            print(f"Usage: {badargs[0]} [python|numpy]", file=sys.stderr)
            sys.exit(1)
    

if __name__ == "__main__":
    def main()
```

so that you can do 

```shell
uv sync --frozen --python=$(uv run scripts/version_bounds.py python)
uv pip install --upgrade numpy==$(uv run scripts/version_bounds.py numpy)
```

scripts/__init__.py
jorenham, line 1: this is not needed now anymore right?

miloth: Yeah, sorry. I was falling asleep on the keyboard during the last commits... That's why I didn't ask for a re-review 😆

jorenham: haha that's the 10x dev life 😆

Issue #261, opened 2024-12-05
:bug: `mdformat-pyproject` not running properly


miloth:
I found the following issue while investigating the task-runners. Running `mdformat . --check` will print files in `.venv` and `.tox`, even though there are settings to exclude them:
```toml
[tool.mdformat]
end_of_line = "lf"
exclude = [".tox/**", ".venv/**", "**/*.m[!d]", "**/*.[!m]d"]
```

This is due to the `mdformat-pyproject` plugin.

PR #262, opened 2024-12-05
➖ remove broken `mdformat-pyproject`

miloth:
- Moved config to `.mdformat.toml` and removed `mdformat-pyproject` as the latter does not support all of the config. Solves #261.
- ~Simplified `poe` command now that the config works.~
- ~Explicitly stating that the dev environment should use Python 3.13, required by `mdformat`'s `exclude` config.~


pyproject.toml
```diff
@@ -107,15 +106,7 @@ rm -rf
...
-mdformat = """
-mdformat
-    CODE_OF_CONDUCT.md
-    CONTRIBUTING.md
-    README.md
-    SECURITY.md
-    tests/README.md
-    codegen/README.md
-"""
+mdformat = "mdformat ."
```
jorenham, line 109: I usually use the lowest supported python version when developing, which can help catch bugs. And since `exclude` requires `python>=3.13`, the markdown files need to be explicitly passed for `mdformat` to work on `python<3.13`. I agree that it's not DRY and ugly, but in this case it's a necessary workaround. 

You could consider submitting a PR to `mdformat` that fixes it; I can't imagine it being all that difficult to implement.

jorenham:
closely related to https://github.com/csala/mdformat-pyproject/issues/4

pyproject.toml
miloth: Another option here would be to remove the `mdformat` dependency (and its plugins) from the dev uv deps, keeping as a separate group:
```toml
[dependency-groups]
mdformat = ["mdformat", "plugins"]
```

Then, change the `poe` command to use a self-contained env made on purpose for the MD check:
```bash
uv run --isolated --group=mdformat --python=3.13 mdformat .
```

jorenham: Hmm, I think I prefer a list of explicit markdown files, as this will slow down stuff like pre-commit, which is bad for DX.

jorenham:
Since it looks like `mkdformat-pyproject` isn't functional at the moment, I think that dropping it is indeed the best opption for now.

Could you rebase, so I can merge?

pyproject.toml
miloth: Do you want this reverted before merging?

CONTRIBUTING.md
```diff
@@ -186,9 +186,12 @@ Ensure you have [`uv`](https://docs.astral.sh/uv/getting-started/installation/)
 Now you can install the project with the dev dependencies:
 
 ```bash
-uv sync --frozen
+uv sync --frozen  --python=3.13
 ```
 
+> [!NOTE]
+> Python 3.13 is required to properly run `mdformat` (see the documentation [note here][mdformat-3.13]).
```
jorenham, line 193:
```suggestion
> Python 3.13 is required to properly run `mdformat` (see the [documentation][mdformat-3.13]).
```

pyproject.toml
jorenham: Yea I think that would be best. And then maybe add a comment that says this is a workaround for the missing mdformat `exclude`with `python<3.13`.

CONTRIBUTING.md
jorenham, line 1: I suppose that these changes here also aren't needed if we keep the `python < 3.13` in place.

pyproject.toml
```diff
@@ -115,7 +114,7 @@ mdformat
     SECURITY.md
     tests/README.md
     codegen/README.md
-"""
+"""  # TODO: This is required because the exclude config only works on Python 3.13+
```
jorenham, line 117: I just raised https://github.com/hukkin/mdformat/issues/489. It could help to link to that here:

```suggestion
"""
# TODO: Remove once `--exclude` can be used on `python<3.13`, see
# https://github.com/hukkin/mdformat/issues/489
```

.mdformat.toml
```diff
@@ -0,0 +1,2 @@
+end_of_line = "lf"
+exclude = [".tox/**", ".venv/**", "**/*.m[!d]", "**/*.[!m]d"]
```
jorenham, line 2: won't `exclude` result in an error on `python<3.13`?

jorenham: ... because if so, then this probably should be commented out until hukkin/mdformat#489 is implemented

jorenham:
should be good to merge after the typo is fixed :)

jorenham (approved):

jorenham:
Thanks @miloth

Issue #101, opened 2024-10-18
🌕 complete `scipy.interpolate`


jorenham:
it uses `Untyped*` ~489~ ~465~ ~376~ ~297~ ~289~ ~221~ ~189~ ~160~ ~128~ ~65~ 43 times

- [x] `dfitpack`
- [x] `fitpack2`
- [x] `fitpack`
- [x] `interpnd`
- [x] `interpolate`
- [x] `ndgriddata`
- [x] `polyint`
- [x] `rbf`
- [x] `_bsplines`
- [x] `_cubic`
- [x] `_fitpack2`
- [x] `_fitpack_impl`
- [x] `_fitpack_py`
- [x] `_interpolate`
- [x] `_ndbspline`
- [x] `_ndgriddata`
- [x] `_pade`
- [x] `_polyint`
- [x] `_rbfinterp`
- [x] `_rbfinterp_pythran`
- [x] `_rbf`
- [x] `_rgi`

berhoel:
The signature for "interpolate._cubic.CubicSpline" is wrong. Argument ``bc_type`` should be
  bc_type:  Literal["not-a-knot","periodic","clamped","natural"]|tuple[tuple[int,float|ArrayLike]|Literal["not-a-knot","clamped","natural"],tuple[int,float|ArrayLike]|Literal["not-a-knot","clamped","natural"]] = "not-a-knot"

jorenham:
> The signature for "interpolate._cubic.CubicSpline" is wrong. Argument `bc_type` should be bc_type: Literal["not-a-knot","periodic","clamped","natural"]|tuple[tuple[int,float|ArrayLike]|Literal["not-a-knot","clamped","natural"],tuple[int,float|ArrayLike]|Literal["not-a-knot","clamped","natural"]] = "not-a-knot"

Rhanks for the reporting this, @berhoel. I have fixed it in #182.

PR #267, opened 2024-12-07
🎨 `interpolate`: remove duplicated function and classes

jorenham:
towards #101 (-39)


jorenham:
this is a classic case of "but it works on my machine"...

PR #285, opened 2024-12-09
:construction_worker: `tox` for cis

miloth:
It still needs a bit of cleanup, this is just a PoC:

- `tox` config in its own file, as it was getting quite long.
- `tox` env cases cover all of the CIs requirement.
- The CIs fully use `tox` environments.

Open issues:
- [ ] Solve `pyright` not working as the last step of the lint job.
- [ ] Why are the `mypy` errors with the minimum version popping up only now?
- [ ] Why the python version does not default to 3.10 in this PR but it does here? https://github.com/miloth/scipy-stubs/pull/4


jorenham:
> * Why the python version does not default to 3.10 in this PR but it does here? [Tox for cis miloth/scipy-stubs#4](https://github.com/miloth/scipy-stubs/pull/4)

maybe it's a caching thing 🤔

jorenham:
> * Why are the `mypy` errors with the minimum version popping up only now?

see #289

jorenham:
the typing errors with old numpy versions have been solved

jorenham:
@wolph this seems to overlap with your PR #286. Do you think that perhaps this could be used to simplify your test matrix 🤔?

jorenham:
superseded by (and merged into) #286

PR #286, opened 2024-12-09
Dynamic testing matrix for combinations of python and numpy versions.

wolph:
Partially custom written, partially GPT generated ;)

Let me know if it needs improvements. At least it finally works.


.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
 
-      - name: install
-        run: uv sync --frozen --python 3.13
-
-      - name: typetest
-        run: uv run poe typetest
+      - name: Install Python Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install scipy
```
jorenham, line 60: the `scipy` version should probably be pinned to the one supported by `scipy-stubs`

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
 
-      - name: stubtest
-        run: uv run poe stubtest
+      - name: Generate Matrix
+        id: set-matrix
+        shell: bash
+        run: |
+          matrix=$(python scripts/generate_matrix.py | jq -c)
```
jorenham, line 66: out of curiosity; what does the `jq -c` do here?

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
       - name: basedpyright
         run: uv run poe pyright
 
-  typetest:
-    timeout-minutes: 5
+  generate-matrix:
+    name: Generate Test Matrix
```
jorenham, line 47: perhaps we should also add a timeout here, just in case

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
+    strategy:
+      matrix: ${{ fromJSON(needs.generate-matrix.outputs.matrix) }}
     steps:
       - uses: actions/checkout@v4
 
-      - name: setup uv
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
```
jorenham, line 81: Maybe I'm missing something, but I thought this wasn't needed anymore because of `uv` 🤔

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
+
+      - name: Install uv
         uses: astral-sh/setup-uv@v4
 
-      - name: install
+      - name: Install Dependencies
```
jorenham, line 86:
```suggestion
      - name: Install dependencies
```

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
         run: |
-          uv sync --frozen --python=$(uv run --frozen scripts/version_bounds.py python)
-          uv pip install --upgrade numpy==$(uv run --frozen scripts/version_bounds.py numpy)
+          uv sync --frozen --python="${{ matrix.python }}"
+          uv pip install numpy=="${{ matrix.numpy }}"
 
-      - name: typetest
+      - name: Run Type Tests
```
jorenham, line 91:
```suggestion
      - name: Run type-tests
```

althugh I'm not sure on the `-` though  🤔

.github/workflows/ci.yml
```diff
@@ -43,40 +43,53 @@ jobs:
...
         run: uv run poe typetest
 
-      - name: stubtest
+      - name: Run Stub Tests
```
jorenham, line 94: `stubtest` is the command name itself, so maybe something like this?

```suggestion
      - name: Run stubtest
```

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
+# !/usr/bin/env python3
```
jorenham, line 1: Can this be removed?

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+import sys
+import urllib.error
+import urllib.request
+from functools import lru_cache
+
+# Constants for caching
+CACHE_SIZE = 128  # Adjust as needed
+
+
+@dataclasses.dataclass(frozen=True, eq=True, order=True)
```
jorenham, line 36: `eq=True` by default, and `slots` isn't

```suggestion
@dataclasses.dataclass(frozen=True, order=True, slots=True)
```

jorenham: Alternatively, you could also consider a `NamedTuple` here. That way, `minor_version` could be a simple slice. But it's also fine like this I suppose 🤷🏻

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def __str__(self) -> str:
+        return f"{self.major}.{self.minor}.{self.micro}"
+
+    @property
+    def minor_version(self) -> tuple[int, int]:
+        return self.major, self.minor
+
+    @classmethod
+    def from_str(cls, version_str: str) -> "Version":
```
jorenham, line 50: It probably won't matter in practice, but the `-> "Version"` could be made theoretically valid by marking the `Version` class as `@final`.

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+        if version_str.count(".") == 1:
+            version_str += ".0"
+        return cls(*map(int, version_str.strip().split(".")))
```
jorenham, line 53: This wekk, `scipy` will release `1.15.0rc1`, which we'll be using to prepare for the upcoming `scipy-stubs==1.15.0.0` release. But it looks like that'll result in a  `ValueError` here.

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Examples:
+        >>> get_minimum_python_version()  # doctest: +ELLIPSIS
+        Version(major=3, ...)
+    """
+    raw_version = importlib.metadata.metadata(package)["Requires-Python"]
+    if "<" in raw_version:
+        raise NotImplementedError("Version specifier with upper bound not yet supported!")
+
+    return Version.from_str(raw_version.replace(">=", "").replace("~=", ""))
```
jorenham, line 71: This looks a lot like `scipy_minimum_python()` in `scripts/version_bounds.py`, so maybe this can be used there, or the other way around, to make it more DRY?

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Examples:
+        >>> get_minimum_version("scipy", "numpy")  # doctest: +ELLIPSIS
+        Version(major=1, ...)
+    """
+    np_minimum_dep = next(
+        req for req in importlib.metadata.requires(package) if req.startswith(dependency) and " extra " not in req
+    )
+    np_minimum_dep = next(ver for ver in np_minimum_dep.split(",") if ">" in ver)
+    return Version.from_str(np_minimum_dep.replace(dependency, "").replace(">=", "").replace(">", ""))
```
jorenham, line 93: And this looks a lot like `scipy_minimum_numpy()` in `scripts/version_bounds.py`, like the one above

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+        if max_version and version > max_version:
+            continue
+
+        versions[version.minor_version] = version
+
+    return sorted(versions.values())
+
+
+@lru_cache(maxsize=CACHE_SIZE)
+def fetch_json(url: str) -> dict | list:
```
jorenham, line 141: `dict` and `list` are missing their type arguments

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+    Examples:
+        >>> data = fetch_json("https://pypi.org/pypi/numpy/json")
+        >>> isinstance(data, dict)
+        True
+    """
+    try:
+        with urllib.request.urlopen(url) as response:  # noqa: S310
+            return json.loads(response.read().decode())
+    except urllib.error.URLError:
+        sys.exit(1)
```
jorenham, line 163: it could help when debugging if this also prints the error message

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+        urllib.error.URLError: If fetching data fails.
+
+    Examples:
+        >>> data = fetch_json("https://pypi.org/pypi/numpy/json")
+        >>> isinstance(data, dict)
+        True
+    """
+    try:
+        with urllib.request.urlopen(url) as response:  # noqa: S310
+            return json.loads(response.read().decode())
```
jorenham, line 161: I believe `json.loads` also accepts `bytes`

```suggestion
            return json.loads(response.read())
```

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+    package_name = "scipy"
+    dependency_name = "numpy"
+
+    min_python_version = get_minimum_python_version(package_name)
+    python_versions = get_python_versions(min_python_version)
+
+    min_package_version = max(
+        get_minimum_version(package_name, dependency_name),
+        # Due to the distutils requirement we cannot use old numpy versions
+        Version(1, 26, 0),
```
jorenham, line 214: `scipy` currently supports `numpy>=1.23.5` 😅

scripts/generate_matrix.py
```diff
@@ -0,0 +1,227 @@
...
+        >>> versions = get_package_versions("numpy", "1.21")
+        >>> "1.21.0" in versions
+        True
+    """
+    data = fetch_json(f"https://pypi.org/pypi/{package_name}/json")
+    releases = data.get("releases", {})
+    versions = {}
+    for version_str in releases:
+        if any(c in version_str for c in ["a", "b", "rc", "dev"]):
+            continue  # Skip pre-releases
```
jorenham, line 187: I'm not sure that's always what we want, given the upcoming `scipy=1.15.0rc1` that we're gonna be using

jorenham (commented):
Thanks for this! I left a couple of questions and suggestions, and many of them are actually about the same thing

scripts/generate_matrix.py
wolph: Yes, we don't really need it in this case.

scripts/generate_matrix.py
wolph: Good plan!

.github/workflows/ci.yml
wolph: Github actions has some shell magic that breaks down from the matrix. The `jq -c` command condenses the json so it doesn't break things. Also in the case of multiline json which can be convenient for debugging :)

.github/workflows/ci.yml
wolph: You're right... force of habit :P
I'm not used to using uv yet

scripts/generate_matrix.py
wolph: I'll add it for now, but it might be useful when differentiating between regular versions and pre-releases

scripts/generate_matrix.py
wolph: True... I assumed we only wanted to test for final releases. I'm explicitly ignoring pre-releases.

scripts/generate_matrix.py
wolph: I forgot to remove that file, it's obsolete since I've moved the functions here.

scripts/generate_matrix.py
wolph: I've added an any, no point in making these type hints more complicated for now :)

wolph (commented):
Addressed all issues, but now it's broken

scripts/generate_matrix.py
jorenham: not needed here, but in case you encounter something like this again, then there's [`optype.json`](https://github.com/jorenham/optype?tab=readme-ov-file#optypejson) that could help with that

.github/workflows/ci.yml
jorenham: > I'm not used to using uv yet

and given that the CI wasn't actually testing the oldest numpy versions before today, apparently neither am I 😛

jorenham:
not sure if relevant here, but ruff supports [PEP 723](https://peps.python.org/pep-0723/) for defining dependecies in standalone scripts:
https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

scripts/generate_matrix.py
```diff
@@ -0,0 +1,200 @@
...
+    Returns:
+        str: The minimum required version of the package.
+
+    Examples:
+        >>> get_minimum_version("scipy", "numpy")  # doctest: +ELLIPSIS
+        Version(major=1, ...)
+    """
+    np_minimum_dep = next(
+        req for req in importlib.metadata.requires(package) if req.startswith(dependency) and " extra " not in req
+    )
```
jorenham, line 72: in case you hate ruff for formatting it like this as well, you can work around it with a `# fmt: skip` pragma thingy, which applies to the entire statement, so e.g.

```suggestion
    np_minimum_dep = next(
        req
        for req in importlib.metadata.requires(package) 
        if req.startswith(dependency)
        and " extra " not in req
    )  # fmt: skip
```

.github/workflows/ci.yml
```diff
@@ -28,55 +28,59 @@ jobs:
...
+      - name: Install uv
         uses: astral-sh/setup-uv@v4
 
-      - name: install
+      - name: Install dependencies
         run: |
-          uv sync --frozen --python=$(uv run --frozen scripts/version_bounds.py python)
-          uv pip install --upgrade numpy==$(uv run --frozen scripts/version_bounds.py numpy)
+          uv sync --frozen --python="${{ matrix.python }}"
+          uv pip install numpy=="${{ matrix.numpy }}"
```
jorenham, line 71: would `$matrix.numpy` also work here 🤔 ?

jorenham: ... or is the `.` a special github actions thingy ?

jorenham:
in case you didn't notice yet; this needs a rebase

scripts/generate_matrix.py
```diff
@@ -57,85 +54,82 @@ def get_minimum_version(package: str, dependency: str) -> Version:
...
+def fetch_json(url: str) -> dict[str, typing.Any] | list[typing.Any]:
     """
     Fetch JSON data from a URL with caching.
 
     Args:
         url (str): The URL to fetch.
 
     Returns:
-        Union[Dict, List]: The parsed JSON data.
+        dict: The parsed JSON data.
```
jorenham, line 128: it's annotated to also being able to return a list 🤔

scripts/generate_matrix.py
```diff
@@ -189,11 +200,22 @@ def main() -> None:
...
-    matrix = {
-        "python": [str(v) for v in python_versions],
-        "numpy": [str(v) for v in package_versions],
-    }
-    json.dump(matrix, fp=sys.stdout)
+    include: list[dict[str, str]] = []
+
+    for python_version in python_versions:
+        for package_version, requires_python in package_versions.items():
+            specifier_set = SpecifierSet(requires_python)
```
jorenham, line 207: swapping the loops would allow `SpecifierSet` to be placed in the outer loop

jorenham (commented):
this matrix script is becoming rather awesome 😮

.github/workflows/ci.yml
```diff
@@ -2,21 +2,20 @@ name: CI
...
   pull_request:
   workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
-env:
-  UV_LOCKED: true
-
```
jorenham, line 15: this removes the needs for `--frozen`, so it should probably be restored

.github/workflows/ci.yml
```diff
@@ -29,92 +28,59 @@ jobs:
...
-      - name: basedpyright
-        run: >
-          uv run --no-sync
-          basedpyright
+      - name: Generate Matrix
+        id: set-matrix
+        shell: bash
+        run: |
+          matrix=$(uv run scripts/generate_matrix.py | jq -c)
+          echo "matrix=$matrix" >> $GITHUB_OUTPUT
```
jorenham, line 54: Maybe `uv pip install packaging` could be replaced with `uv add --script`?
https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies

or alternatively, `uv run --with packaging ...`?

.github/workflows/ci.yml
```diff
@@ -29,92 +28,59 @@ jobs:
...
-          uv run --no-sync
-          stubtest
-          --mypy-config-file=pyproject.toml
-          --allowlist=.mypyignore
-          --ignore-unused-allowlist
-          scipy
+        run: uv run poe mypy
+
+      - name: basedpyright
+        run: uv run poe pyright
```
jorenham, line 86: The `uv run poe` for mypy and stubtest uses `--isolated`, which creates a new venv, and therefore ignores the `uv pip install numpy==...`.
So that's the reason for all that ugly `uv run --no-sync` stuff

scripts/generate_matrix.py
```diff
@@ -0,0 +1,273 @@
...
+    stable: bool
+    release_url: str
+    files: list[FileInfo]
+
+
+class PackageVersions(typing.TypedDict, total=False):
+    releases: dict[str, list[FileInfo]]
+
+
+@lru_cache(maxsize=CACHE_SIZE)  # type: ignore[no-any-expr]
```
jorenham, line 40: wouldn't a `functools.cache` be enough in this case?

scripts/generate_matrix.py
```diff
@@ -0,0 +1,273 @@
...
+        if max_version and version > max_version:
+            continue
+
+        versions[version.major, version.minor] = version
+
+    return sorted(versions.values())
+
+
+@lru_cache(maxsize=CACHE_SIZE)  # type: ignore[no-any-expr]
+def fetch_json(url: str) -> object:
```
jorenham, line 146: returning `Any` avoids having to `typing.cast` in `get_available_package_versions` (`Any` is still evil, though)

scripts/generate_matrix.py
```diff
@@ -0,0 +1,273 @@
...
+        if requires_python is None:
+            # Skip versions without 'requires_python'
+            continue
+
+        key = (version.major, version.minor)
+        # Update to latest version within the minor version series
+        if key not in latest_versions or version > latest_versions[key][0]:
+            latest_versions[key] = (version, requires_python)
+
+            # Extract the versions and requires_python from the latest_versions dict
```
jorenham, line 214:
```suggestion
    # Extract the versions and requires_python from the latest_versions dict
```

scripts/generate_matrix.py
```diff
@@ -0,0 +1,273 @@
...
+                    break
+
+        if requires_python is None:
+            # Skip versions without 'requires_python'
+            continue
+
+        key = (version.major, version.minor)
+        # Update to latest version within the minor version series
+        if key not in latest_versions or version > latest_versions[key][0]:
+            latest_versions[key] = (version, requires_python)
```
jorenham, line 212:
```suggestion
            latest_versions[key] = version, requires_python
```

scripts/generate_matrix.py
```diff
@@ -0,0 +1,273 @@
...
+                include.append(
+                    {
+                        "python": str(python_version),
+                        DEPENDENCY_NAME: str(package_version),
+                    }
+                )
+
+    matrix = {"include": include}
+
+    json.dump(matrix, indent=4, fp=sys.stdout)
```
jorenham, line 269: is the indent needed?

jorenham, line 269: the assignment to `matrix` feels a bit redundant here I guess 🤷🏻

jorenham (commented):
I don't think the installed numpy versions are actually used by mypy and stubtest this way, as I explained in the relevant comments.

Oh and feel feel to ignore my neurotic stylistic suggestions 😛

scripts/generate_matrix.py
wolph: I agree, it was really convenient for debugging purposes (i.e. `rich.print(matrix)`)

scripts/generate_matrix.py
wolph: YES! Why do many most formatters/linters/etc. love brackets so much? Are they secretly lisp-lovers? Could they be emacs users?!111cos(0)?

scripts/generate_matrix.py
wolph: blame GPT-o1... wasn't me

scripts/generate_matrix.py
wolph: It seems to make mypy unhappy though... so it doesn't help :(

scripts/generate_matrix.py
wolph: image

once again... I blame GPT, I didn't write that code I just told it to use caching ;)

scripts/generate_matrix.py
jorenham: yea, `based{mypy,pyright}` declared a holy war against `Any` (and therefore gradual typing as a whole, for which they recommend their "baseline" feature as an alternative), so you'd have to ignore some errors in that case, so perhaps a `cast` is indeed cleaner

scripts/generate_matrix.py
jorenham: 🤣

jorenham: but `ruff format` doesn't add these `(...)` right 🤔?

wolph:
It's finally done! And it includes @miloth his tox implementation.

tox.toml
```diff
@@ -0,0 +1,79 @@
...
+
+[env.basedpyright-tests]
+description = "basedpyright tests"
+dependency_groups = ["typecheck"]
+commands = [["basedpyright", "tests"]]
+
+[env.stubtest]
+description = "stubtest"
+dependency_groups = ["typecheck"]
+commands = [["uv", "run", "--frozen", "--no-editable", "--isolated", "--refresh-package=scipy-stubs", "stubtest", "--mypy-config-file=pyproject.toml", "--allowlist=.mypyignore", "--ignore-unused-allowlist", "scipy"]]
```
jorenham, line 63: https://www.youtube.com/watch?v=6-1Ue0FFrHY

... so, maybe linebreaks 😛 ?

.github/workflows/ci.yml
```diff
@@ -29,92 +28,69 @@ jobs:
...
-        with:
-          enable-cache: true
-          python-version: "3.13"
 
-      - name: install
-        run: uv sync --no-editable
+      - name: Install dependencies
+        run: |
+          uv sync --frozen
+          uv pip install packaging
```
jorenham, line 47: packaging is a dev dep now, so I dont think this is needed

.github/workflows/ci.yml
```diff
@@ -29,92 +28,69 @@ jobs:
...
         uses: astral-sh/setup-uv@v4
-        with:
-          enable-cache: true
-          python-version: "3.13"
 
-      - name: install
-        run: uv sync --no-editable
+      - name: Install dependencies
+        run: |
+          uv sync --frozen
```
jorenham, line 46: how about `--locked`, so we enforce an up-to-date lockfile?

.github/workflows/ci.yml
```diff
@@ -29,92 +28,69 @@ jobs:
...
-          --allowlist=.mypyignore
-          --ignore-unused-allowlist
-          scipy
-
-  typetest-oldest-supported-numpy:
-    timeout-minutes: 5
+  typetest:
+
+    needs: generate-matrix
+    timeout-minutes: 20
```
jorenham, line 59: maybe 20 is overkill; it currently runs in ~1 min

.github/workflows/ci.yml
```diff
@@ -29,92 +28,69 @@ jobs:
...
     steps:
       - uses: actions/checkout@v4
 
-      - name: setup uv
+      - name: Install uv
         uses: astral-sh/setup-uv@v4
         with:
           enable-cache: true
-          python-version: "3.10"
+          python-version: ${{ matrix.python }}
```
jorenham, line 71: doesn't this remove the need for the explicit `--python=...` later on?

jorenham (commented):
it might be a good idea to explicitly print the numpy version from within the tox env, so that we can validate that it's actually using the correct numpy version

.github/workflows/ci.yml
wolph: Maybe? I'll test

wolph:
> it might be a good idea to explicitly print the numpy version from within the tox env, so that we can validate that it's actually using the correct one

I think it accidentally already does that:
image

jorenham (approved):

jorenham:
Thanks @wolph and @miloth !

PR #289, opened 2024-12-09
💚 fix `typetest-oldest-supported-numpy` CI job

jorenham:
apparently mypy was still using the latest numpy because it doesn't understand editable projects and has to be run in an isolated uv env which doesn't include the `uv pip install`'ed numpy version or the correct python version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,13 +70,31 @@ jobs:
       - name: setup uv
         uses: astral-sh/setup-uv@v4
 
-      - name: install
-        run: |
-          uv sync --frozen --python=$(uv run --frozen scripts/version_bounds.py python)
-          uv pip install --upgrade numpy==$(uv run --frozen scripts/version_bounds.py numpy)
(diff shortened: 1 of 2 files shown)


jorenham:
@miloth @wolph any thoughts on this? It's horribly ugly, but at least it actually works now 😅

jorenham:
I'm gonna merge this as a temporary fix that will be properly fixed by #285

wolph:
> @miloth @wolph any thoughts on this? It's horribly ugly, but at least it actually works now 😅 

Won't my pull request be able to automatically fix this because it runs the separate environments already?

jorenham:
> Won't my pull request be able to automatically fix this because it runs the separate environments already?

Well, no, because otherwise the CI would have been failing, as there are a bunch of `numpy<1.24` errors after all (see e.g. https://github.com/jorenham/scipy-stubs/pull/292) 😅 

And this one also didn't fix it, but https://github.com/jorenham/scipy-stubs/pull/290 did 🤷🏻

PR #363, opened 2024-12-21
🔧 no build isolation for `uv`

jorenham:


jorenham:
oh nevermind, this breaks stuff

Issue #447, opened 2025-02-05
Incorrect type for `prominence` argument to `scipy.signal.find_peaks()`


drewsilcock:
If you look at the function `scipy.signal.find_peaks()` found in [`scipy.signal._peak_finding.py` line 729](https://github.com/scipy/scipy/blob/main/scipy/signal/_peak_finding.py#L729), the docstring indicates the following about the argument `prominence`:

```
prominence : number or ndarray or sequence, optional
        Required prominence of peaks. Either a number, ``None``, an array
        matching `x` or a 2-element sequence of the former. The first
        element is always interpreted as the  minimal and the second, if
        supplied, as the maximal required prominence.
```

However, this argument is [currently stubbed with](https://github.com/scipy/scipy-stubs/blob/master/scipy-stubs/signal/_peak_finding.pyi#L70):

```
    prominence: onp.ToFloatND | Sequence[onp.ToFloat | None] | None = None,
```

As such, it's missing the "just a plain number" option from the typing.

Proposed change:

```
    prominence: onp.ToFloat | onp.ToFloatND | Sequence[onp.ToFloat | None] | None = None,
```

PR #448, opened 2025-02-05
Bugfix: Add plain number option for find_peaks argument prominence

drewsilcock:
See #447 for details.

### DIFF
--- a/scipy-stubs/signal/_peak_finding.pyi
+++ b/scipy-stubs/signal/_peak_finding.pyi
@@ -72,7 +72,7 @@ def find_peaks(
     height: onp.ToFloatND | Sequence[onp.ToFloat | None] | onp.ToFloat | None = None,
     threshold: onp.ToFloatND | Sequence[onp.ToFloat | None] | None = None,
     distance: onp.ToFloat | None = None,
-    prominence: onp.ToFloatND | Sequence[onp.ToFloat | None] | None = None,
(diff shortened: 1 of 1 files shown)


jorenham:
Thanks! 

Any idea whether this also affects the returned `_FindPeaksResultsDict` key `prominences`?

jorenham (approved):

drewsilcock:
Thanks for the quick response. I don't believe this affects `_FindPeaksResultsDict` key `prominences` – the docstring for the `find_peaks()` function says:

```
* 'prominences', 'right_bases', 'left_bases'
              If `prominence` is given, these keys are accessible. See
              `peak_prominences` for a description of their content.
```

While the `peak_prominences()` function docstring says:

```
    Returns
    -------
    prominences : ndarray
        The calculated prominences for each peak in `peaks`.
    left_bases, right_bases : ndarray
        The peaks' bases as indices in `x` to the left and right of each peak.
        The higher base of each pair is a peak's lowest contour line.
```

While I've not run the code and examined the runtime type that's returned, this seems clear that `_FindPeaksResultsDict["prominences"]` is of type ndarray as that's what `peak_prominences()` says is the type.

jorenham:
Welcome aboard, Drew 🎉

drewsilcock:
Thanks very much, Joren 🙂

PR #451, opened 2025-02-05
Bugfix: Make `optimizer` argument optional in `scipy.stats.rv_continuous.fit()`

drewsilcock:
See https://github.com/scipy/scipy-stubs/issues/450 for details.

### DIFF
--- a/scipy-stubs/stats/_distn_infrastructure.pyi
+++ b/scipy-stubs/stats/_distn_infrastructure.pyi
@@ -805,7 +805,7 @@ class rv_continuous(_rv_mixin, rv_generic):
         optimizer: Callable[
             [_FloatND, tuple[float, ...], tuple[float, ...], bool],
             tuple[onp.ToFloat, ...],
-        ],
+        ] | None = None,
         method: _FitMethod = "MLE",
         **kwds: onp.ToFloat,
     ) -> tuple[_Float, ...]: ...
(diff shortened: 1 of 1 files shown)


drewsilcock:
Not sure what's going on with the CI tests – it seems to be failing on the "setup uv" step which I am unlikely to have broken (although I wouldn't put it past me).

scipy-stubs/stats/_distn_infrastructure.pyi
```diff
@@ -805,7 +805,8 @@ class rv_continuous(_rv_mixin, rv_generic):
         optimizer: Callable[
             [_FloatND, tuple[float, ...], tuple[float, ...], bool],
             tuple[onp.ToFloat, ...],
-        ],
+        ]
+        | None = None,
```
jorenham, line 809: I just tried with `None`, and that seems to work (although I'm not sure why). But since it defaults to `optimize.fmin`, I think it'd be better do use `...` as default here:

```suggestion
        | None = ...,
```

jorenham:
> Not sure what's going on with the CI tests – it seems to be failing on the "setup uv" step which I am unlikely to have broken (although I wouldn't put it past me).

The CI failures are unrelated. 
They're caused by the latests `3.13.2` and `3.12.9` Python releases, which are not yet installable through `uv` apparently. I'm looking into a workaround as we speak.

jorenham (approved):

drewsilcock:
Thanks Joren

jorenham:
Thanks Drew!

PR #470, opened 2025-03-06
🐛 `spatial.transform`: accept proper Euler angles in `Rotation.as_euler`

KiSchnelle:
Current definition

>  _RotAxisSeq: TypeAlias = L["xyz", "xzy", "yzx", "yxz", "zxy", "zyx", "XYZ", "XZY", "YXZ", "YZX", "ZXY", "ZYX"]

only includes Tait-Bryan sequences.

Documentation at:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.as_euler.html

shows its viable to use proper Euler also.

> [0, 180] degrees if first and third axes are the same (like zxz)

### DIFF
--- a/scipy-stubs/spatial/transform/_rotation.pyi
+++ b/scipy-stubs/spatial/transform/_rotation.pyi
@@ -17,31 +17,11 @@ _Float2D3D: TypeAlias = _Float2D | _Float3D
 _RotOrder: TypeAlias = L["e", "extrinsic", "i", "intrinsic"]
 _RotGroup: TypeAlias = L["I", "O", "T", "D", "Dn", "C", "Cn"]
 _RotAxisSeq: TypeAlias = L[
-    "xyz",
-    "xzy",
-    "yxz",
-    "yzx",
-    "zxy",
-    "zyx",
-    "xyx",
-    "xzx",
-    "yxy",
-    "yzy",
-    "zxz",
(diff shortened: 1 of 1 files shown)


scipy-stubs/spatial/transform/_rotation.pyi
```diff
@@ -16,7 +16,32 @@ _Float2D3D: TypeAlias = _Float2D | _Float3D
...
+    "YZX",
+    "ZXY",
+    "ZYX",
+    "XYX",
+    "XZX",
+    "YXY",
+    "YZY",
+    "ZXZ",
+    "ZYZ",
+]
```
jorenham, line 44: I saw the commit message, and I agree that `ruff format` is being kinda annoying here. So you could disable the formatter by adding a `]  # fmt: skip` here, which would allow you to group the literal values or something.

jorenham:
Ah I think I understand it now; there can be duplication, but only if the duplicates are on "the outside".

scipy-stubs/spatial/transform/_rotation.pyi
KiSchnelle: I edited it, I really like ruff but I think here it's easier to read grouped into intrinsic/extrinsic Tait-Bryan and proper Euler. Thx

PR #476, opened 2025-03-11
🐛 `scipy.interpolate`: accept scalar input in `BSpline.__call__`

### LINKED ISSUE #473: Bug: Support for scalars in typing of array_likes
At the example of The first argument of `Bspline.__call__`

https://github.com/scipy/scipy-stubs/blob/0ed2f4eecb51866d944bc2bb44a727f4c5fe01cf/scipy-stubs/interpolate/_bsplines.pyi#L40

- [docs say](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.__call__.html#scipy.interpolate.BSpline.__call__) "x : array_like", and according to [numpy](https://numpy.org/doc/stable/glossary.html#term-array_like) this also includes scalars
- the implementation also works when calling with a scalar
- but type checkers complain because it is typed as `onp.ToFloatND`, which [according to docs](https://github.com/jorenham/optype?tab=readme-ov-file#array-likes) "don't accept "bare" scalar types"
  `Argument of type "float" cannot be assigned to parameter "x" of type "ToFloatND"`

Is this a bug, or a limitation we cannot work around. If the second, the statement "works with all array-likes" from the README should be clarified.

jorenham:
closes #473

### DIFF
--- a/scipy-stubs/interpolate/_bsplines.pyi
+++ b/scipy-stubs/interpolate/_bsplines.pyi
@@ -9,10 +9,14 @@ from scipy.sparse import csr_array
 
 __all__ = ["BSpline", "make_interp_spline", "make_lsq_spline", "make_smoothing_spline"]
 
+###
+
 _Extrapolate: TypeAlias = Literal["periodic"] | bool
 _BCType: TypeAlias = Literal["not-a-knot", "natural", "clamped", "periodic"]
 _LSQMethod: TypeAlias = Literal["qr", "norm-eq"]
 
+# TODO(jorenham): Change bound to `float64 | complex128`
+# https://github.com/scipy/scipy-stubs/issues/474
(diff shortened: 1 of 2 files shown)


ev-br:
While I've no idea about the typing side of it, from the scipy.interpolate behavior you might want to do something similar to PPoly.__call__ and BPoly.__call__. They have a similar 0D or higher-D return value, depending on self.c.ndim. For these, I belive, `self.c` is at least 2D, though.

jorenham:
> While I've no idea about the typing side of it, from the scipy.interpolate behavior you might want to do something similar to PPoly.**call** and BPoly.**call**. They have a similar 0D or higher-D return value, depending on self.c.ndim. For these, I belive, `self.c` is at least 2D, though.

I've addressed this in https://github.com/scipy/scipy-stubs/pull/477 :)

PR #477, opened 2025-03-11
🐛✨ `scipy.interpolate`: improved support for complex splines

### LINKED ISSUE #474: `scipy.interpolate.BSpline` type-param should be either `float64` or `complex128`
It currently always defaults to `np.floating`, independent of the value of `c` passed to `__init__`. 
But since `_get_dtype` is used to, well, get the dtype, which either returns `float64` or `complex128`, the type-parameter should instead either assume `float64` or `complex128`:
https://github.com/scipy/scipy/blob/0f1fd4a7268b813fa2b844ca6038e4dfdf90084a/scipy/interpolate/_bsplines.py#L21-L26

jorenham:
closes #474

### DIFF
--- a/scipy-stubs/interpolate/_bsplines.pyi
+++ b/scipy-stubs/interpolate/_bsplines.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Generic, Literal, TypeAlias, TypeVar
+from typing import Any, Generic, Literal, TypeAlias, TypeVar, overload
 from typing_extensions import Self
 
 import numpy as np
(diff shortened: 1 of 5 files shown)


scipy-stubs/interpolate/_bsplines.pyi
```diff
@@ -11,66 +11,109 @@ __all__ = ["BSpline", "make_interp_spline", "make_lsq_spline", "make_smoothing_s
...
+    def from_power_basis(cls, pp: CubicSpline[_CT_co], bc_type: _BCType = "not-a-knot") -> Self: ...
+
+    #
     @classmethod
     def construct_fast(
         cls,
-        t: onp.Array1D[np.float64],
-        c: onp.Array[onp.AtLeast1D, _SCT_co],
+        t: onp.ArrayND[np.float64],
+        c: onp.ArrayND[_CT_co],
```
ev-br, line 110: `construct_fast` gets exactly the same arguments as `__init__`, should probably be typed the same?

scipy-stubs/interpolate/_bsplines.pyi
```diff
@@ -11,66 +11,109 @@ __all__ = ["BSpline", "make_interp_spline", "make_lsq_spline", "make_smoothing_s
 
 ###
 
+_CT_co = TypeVar("_CT_co", bound=np.float64 | np.complex128, default=np.float64, covariant=True)
```
ev-br, line 14: does this exclude float32? If yes, `BSpline.__init__` accepts it (and upcasts to float64; ditto for complex64 -> complex128); same for `PPoly.__init__` and `BPoly.__init__` (this comment applies everywhere unless noted explicitly)

ev-br (commented):
A couple of hopefully helpful comments only.
Sorry, my knowledge of the typing syntax is not up to par. 
Can help explaining the API intentions if you tell me specifically what to look at and translate from annotations to natural language.
(a reverse LLM, as it were).

scipy-stubs/interpolate/_bsplines.pyi
jorenham: It skips the conversion and validation steps, so it doesn't convert array-likes to arrays, for example. That's why `t` and `c` are annotated to have (almost) the same type as their corresponding attributes.

scipy-stubs/interpolate/_bsplines.pyi
jorenham: `__init__` still accepts any array-like that can be converted to `float64` or `complex128`. The constructor functions and methods use `@overload` to "select" either `float64` or `complex`, based on whether the input is "float-like" or "complex-like". If it can't be determined (e.g. when passing an array type of `np.inexact`), it'll fall back to `Any`.

jorenham:
> Sorry, my knowledge of the typing syntax is not up to par. 

No worries; I was mostly wondering whether I missed anything. And now that I look at it, I actually did miss the spline classes in `_interpnd` and `_nbdspline` 😅

jorenham:
This PR is getting a bit too messy actually, I'll split it up into one for the complex/float stuff, and one for the remaining rejected scalar input to `__call__`

jorenham:
well, I believe that's all of them

jorenham:
Do you want to take another look @ev-br or you ok with merging?

ev-br:
I trust you on typing matters, so don't wait for me :-).
All I'm to say is all 1D and ND interpolators indeed should accept complex inputs (modulo Akima1D and PCHIP, which do not, for algorithmic reasons). 
ND interpolators do not accept scalars.

PR #523, opened 2025-05-21
📌 require `python>=3.11`

jorenham:


jorenham:
This restriction only applies to the `scipy-stubs` versions that support `scipy>=1.16`, and if you use Python 3.10, you also can't use `scipy>=1.16`. If that's the case, then you can use e.g. `scipy-stubs=1.15.3.0`, `1.14.1.6`, or some other version whose prefix matches the installed `scipy` version. 

And for what it's worth, librosa, which also supports 3.9, pinned it as `scipy-stubs; python_version >= '3.10'`: 
https://github.com/librosa/librosa/blob/5981a32a9d6fe6c103688afc7161c726859e9695/setup.cfg#L113

lucascolley:
oh, good shout, sorry for the noise.

lucascolley:
I think things will be fine once we have a release on conda-forge, just an annoying interaction between Pixi and PyPI deps with an optional dependency with a restricted Python range

jorenham:
Anything I can help with?

PR #528, opened 2025-05-22
👽️ scipy 1.16 change for `interpolate.RectBivariateSpline`

jorenham:
A `maxit=20` constructor parameter has been added to the `interpolate.RectBivariateSpline`  class.


tylerjereddy:
This is from a bug fix PR (https://github.com/scipy/scipy/pull/22433), so per the usual convention I can probably get away with not giving it a line the relnotes (though for API changes there may be some merit to mentioning it--for now I'll probably let myself off the hook for the explicit bug fixes, at least for now, but anything not a bug fix you've mentioned I've tried to address)

PR #561, opened 2025-05-26
👽️ scipy 1.16 additions for `spatial.transform._rotation`

jorenham:
A new public function has been added to the private `scipy.spatial.transform._rotation` module, `compose_quat`, with signature `(f64[:, :], f64[:, :]) -> f64[:, :]`.

### DIFF
--- a/.mypyignore-todo
+++ b/.mypyignore-todo
@@ -1,6 +1,6 @@
 scipy.spatial.transform.RigidTransform
 scipy.spatial.transform.__all__
-scipy.spatial.transform._rotation.compose_quat
+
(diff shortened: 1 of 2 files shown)


tylerjereddy:
I don't see any tests for `compose_quat` added in https://github.com/scipy/scipy/pull/22267, so I suspect I may be "ok" in considering it informally private for release notes purposes. I'll copy @scottshambaugh just in case though.

scottshambaugh:
Yeah, `compose_quat` is not meant to be part of the public API, composition is done with the `__mul__` operator. It's an alias for the cdef `_compose_quat` function so that can be imported elsewhere.

PR #566, opened 2025-05-27
👽️ scipy 1.16 new class `stats._distribution_infrastructure`

jorenham:
The new `UnivariateDistribution` and `DiscreteDistribution` classes have been added to the private `scipy.stats._distribution_infrastructure` module.


jorenham:
> I'm almost certain that this is covered by the general relnotes language surrounding:
> 
> > `scipy.stats.make_distribution` was extended to work with existing ...
> 
> In the `stats` section

Ah I see that `scipy.stats` doesn't export them, so this change is indeed irrelevant for the release notes

PR #591, opened 2025-05-29
⬆️ mypy 1.16.0

jorenham:
This fixes new `stubtest` errors, and removes a bunch of `# type: ignore`'s for now-resolved false positive mypy errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
           --no-editable
           basedpyright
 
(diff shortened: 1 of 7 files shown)


lucascolley:
what is the rationale behind dropping basedmypy?

jorenham:
> what is the rationale behind dropping basedmypy?

Because it doesn't include the latest mypy changes. And unless I'm mistaken, I believe that @KotlinIsland does not have the bandwidth to work on it.

PR #616, opened 2025-06-16
✅ `differentiate`: type-tests for `derivative`, `jacobian`, and `hessian`

jorenham:


jorenham:
> * Add a test assertion for `res_hes.nit` to validate the iteration count on the HessianResult.

`.nit` doesn't exist at runtime for `hessian`

PR #624, opened 2025-06-17
✨ `sparse`: improved integer matrix/array method annotations

### LINKED ISSUE #403: overload the `sparse._data._data_matrix` methods with concrete return types

jorenham:
This affects the following `scipy.sparse` types:

- `bsr_matrix`, `bsr_array`
- `coo_matrix`, `coo_array`
- `csc_matrix`, `csc_array`
- `csr_matrix`, `csr_array`
- `dia_matrix`, `dia_array`

and the following methods:

- `sqrt`
- `expm1`
- `log1p`
- `sin`
- `sinh`
- `tan`
- `tanh`
- `arcsin`
- `arcsinh`
- `arctan`
- `arctanh`
- `deg2rad`
- `rad2deg`

When the sparse array/matrix has an integer dtype, the return types of these methods will is now specialized for both the "outer" sparse array/matrix type, and the "inner" generic scalar (and shape) type. So type-checkers will now infer `csc_array([[1, 2], [3, 4]]).sqrt()` as `csc_array[np.float64]`.

This behaviour is verified using type-tests for all 8 sparse array- and matrix types.

---

closes #403


jorenham:
Unprompted flex: This PR took me 1 hour and 5 minutes 😎

PR #642, opened 2025-06-22
📌 require `scipy>=1.16.0<1.17`

jorenham:
SciPy 1.16.0 changed the `scipy.optimize._lbfgsb_py.fmin_l_bfgs_b` default parameter value of `disp` from `-1` to a `_NoValue` sentinel object. This is technically a breaking change, and resulted in 4 new stubtest errors. I have ignored them for now, and will address them in a follow-up


tylerjereddy:
Thanks for keeping me in the loop.

> This is technically a breaking change

I think I reviewed @ilayn's PR that made this change. I believe that from a code behavior standpoint it is backwards compatible because the two args that got switched to `_NoValue` don't get used anyway (i.e., they are basically no-ops at this point). As you probably know, the `_NoValue` is basically a plain sentinel `object` used to detect if the user is still trying to (pointlessly) use one of those args, and if detected the deprecation gets emitted.

That said, let's see if Ilhan has anything to add/correct me on.

I suppose you might be arguing that from a typing perspective this is a break--I don't think I'd argue against that, but I believe that may be allowed as long as the function behavior itself is backwards compatible.

jorenham:
> I suppose you might be arguing that from a typing perspective this is a break

Here's indeed a minor violation of the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) here. But I agree that it's pretty unlikely that this will cause any real issues in practice. 

Besides, every change is a breaking change, according to [Hyrum's Law](https://www.hyrumslaw.com/): 

> With a sufficient number of users of an API,
> it does not matter what you promise in the contract:
> all observable behaviors of your system
> will be depended on by somebody.

So yea, I wouldn't worry about this too much haha

ilayn:
My bad, apologies for that. I didn't remember that you are tracking the types of the kwargs.

`_NoValue` is our "poor-man's `NULL` pointer". If it's not `NULL` then user passed something which they should not have because warnings would turn to errors in a few versions.

jorenham:
> `_NoValue` is our "poor-man's `NULL` pointer". If it's not `NULL` then user passed something which they should not have because warnings would turn to errors in a few versions.

I've seen it in a couple of other places as well. It's pretty difficult to annotate (since `spam: object` accepts *everything*), but I figured out a (kinda hacky) way around that, so it's not much of a problem.
Although I suppose that a dedicated sentinel type like numpy's `numpy._NoValue: numpy._globals._NoValueType` would be easier to type, and wouldn't require typing trickery like this. But for end-users this wouldn't matter at all, so it's not all that important.

PR #719, opened 2025-07-07
🏷️ `sparse`: added stubs for all formats in `kron` and `kronsum`

JulVandenBroeck:
Hi again!

Now that I'm familiar with the PR process, I'm going over all the "# type: ignore" lines in my project and I noticed room for improvement in the stubs for the `kron` and `kronsum` functions.
I added a stub for each different format, so that the end result is of the correct type. Furthermore, since these functions only allow 2D array-like objects, I annotated the input arrays to be 2D by adding `, tuple[int, int]` to the type alias.
I included extra tests as well, which seem to pass.

Let me know your thoughts!

Thanks in advance,
Jul


jorenham:
> Do you know of any better (perhaps programmatic) way to define overloads? Would be cool to gather the types in lists and define overloads in a loop...

Something like that is pretty high on my wishlist. But unfortunately, apart from code-generation, that's not something we can express in Python's current type system.

jorenham:
Nice improvement; thanks!

PR #733, opened 2025-07-09
:label: `sparse`: overhauled stubs for `diags_array`, `eye_array`, and their matrix equivalents

JulVandenBroeck:
Hello!

I improved support for the different sparse formats in `diags`, `spdiags`, `diags_array`, `identity`, `eye`, and `eye_array`. 

Additionally, I made the stubs of the`diags_array` function more versatile, so that it supports Python sequences a little better (see the extra tests). To achieve this I needed the `optype.numpy.AnyFloat64DType`, but without the `None` option, which I reimplemented in the `_construct.pyi` file.
Is there already such a type? Alternatively, is it too late to remove `None` and alter all uses to include ` | None`? I saw that this is the case in most places already. 

Let me know!

Kind regards,
Jul


scipy-stubs/sparse/_construct.pyi
```diff
@@ -7,6 +7,10 @@ import numpy.typing as npt
 import optype.numpy as onp
 import optype.numpy.compat as npc
 import optype.typing as opt
+from optype.numpy._any_dtype import f_cls
```
jorenham, line 10: if you `import optype as op`, then this private type alias can be replaced with `type[op.JustFloat]`

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +42,16 @@ __all__ = [
...
 _T = TypeVar("_T")
 _SCT = TypeVar("_SCT", bound=_Numeric, default=Any)
 _SCT0 = TypeVar("_SCT0", bound=_Numeric)
 _ShapeT = TypeVar("_ShapeT", bound=tuple[int, *tuple[int, ...]], default=tuple[Any, ...])
 
-_Numeric: TypeAlias = npc.number | np.bool_
-
-_ToArray1D: TypeAlias = Seq[_SCT] | onp.CanArrayND[_SCT]
-_ToArray2D: TypeAlias = Seq[Seq[_SCT | int] | onp.CanArrayND[_SCT]] | onp.CanArrayND[_SCT]
+_ToArray1D: TypeAlias = Seq[int | float | complex | _SCT] | onp.CanArrayND[_SCT]
```
jorenham, line 53: Although I hate these `int <: float <: complex` "type promotion rules", there are several practical reasons for why it's best to instead write `int | float | complex` as `complex` (even though it's technically wrong).

But that being said, I believe that this is exactly equivalent to `onp.ToArray1D[complex, _SCT]`

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +42,16 @@ __all__ = [
...
+_ToArray2D: TypeAlias = Seq[_ToArray1D[_SCT]] | onp.CanArrayND[_SCT]
```
jorenham, line 54: Although not exactly equivalent (this way is broader), you might wanna consider writing this as:

```suggestion
_ToArray2D: TypeAlias = onp.ToArray2D[complex, _SCT]
```

scipy-stubs/sparse/_construct.pyi
```diff
@@ -87,212 +91,1180 @@ class _DataSampler(Protocol):
...
-@overload  # diagonals: <known>, dtype: None = ..., format: {"dia", None} = ...
-def diags_array(  # type: ignore[overload-overlap]
+# DIA
+@overload  # diagonals: <known>, format: {"dia", None} = ..., dtype: None
+def diags_array(
     diagonals: _ToArray1D[_SCT] | _ToArray2D[_SCT],
     /,
     *,
-    offsets: onp.ToInt | onp.ToInt1D = 0,
+    offsets: onp.ToInt | onp.ToIntStrict1D = 0,
```
jorenham, line 100: Since the `offsets` all have the same type (right?), a type alias might help.

scipy-stubs/sparse/_construct.pyi
```diff
@@ -87,212 +91,1180 @@ class _DataSampler(Protocol):
 ###
 
 # NOTE: The `overload-overlap` mypy errors are false positives.
-@overload  # diagonals: <known>, dtype: None = ..., format: {"dia", None} = ...
-def diags_array(  # type: ignore[overload-overlap]
+# DIA
+@overload  # diagonals: <known>, format: {"dia", None} = ..., dtype: None
+def diags_array(
```
jorenham, line 96: I count 82 `diags_array` overloads. I know that at least pyright *used to* have a hard limit of 64 overloads, but recently increased this to 1024 (https://github.com/microsoft/pyright/blob/8b02aa84ec916697e76d5ee69299a89d555ec98c/packages/pyright-internal/src/analyzer/constraintTracker.ts#L16-L24). I don't know what the mypy limit is, but seeing as pyright's design choices are often in line with mypy's, I wouldn't be surprised if mypy currently has a 64 overload limit (but there's a good chance I'm wrong though). 

Either way, in general I've been trying to keep to no. of overloads below 64. Because besides those (semi-alleged) type-checker limits, the no. of overloads can become a performance bottleneck (the overload "constraint solving" algo's that type-checkers use are exponential (or even factorial) in time complexity). When I, for example, removed a bunch of redundant overloads for `numpy.dtype.__new__`, the runtimes of both pyright and mypy went down by >10s when ran on the numpy type-tests (I believe it was closer to 20s, but I don't have the numbers at hand, so I'm being conservative here).

Anyway, to make a long story ~yet a bit longer~ short: If you see any way to reduce the no. of overloads here, then please do :)

jorenham: the 77 overloads of `eye` can also do with some pruning, btw (and I suppose it's fine to be a bit more aggressive here, given that it's legacy `*_matrix` stuff)

jorenham:
> I wanted to support an input of type `list[int | float]` or `list[list[int | float]]` (`bool` gives an error in `diags_array`), so that it yields an output of type `dia_array[np.float64]`, which is always the case even for lists with only integers.
> However, as soon as there is one complex number in any of the lists, I want to output `dia_array[np.complex128]`. The problem is that `int` and `float` are treated as subtypes of `complex`, and I was unable to fix this without getting errors about overlapping overloads.

I think the issue is the invariance of `list`. The standard workaround for that is to use the covariant `collections.abc.Sequence`, but that's not always appropriate. Alternatively, you could use a "free" `T = TypeVar("T", int, float)` or something, that you *only* use in `list[T]` (but I didn't try this out myself yet, so I'm not 100% certain it will work).

jorenham:
> I accidentally pushed a VSCode setting, sorry about that. Should be good to go now :)

If you think it can help other VSCode users (like me) as well, feel free to open a separate PR for it.

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: Nice! `onp.ToArray1D[complex, _SCT]` is almost exactly the same, with the addition of 0D numpy arrays. 

After replacing this, Pyright claims that several overloads are redundant now, which seems to be confirmed by the tests still passing.

JulVandenBroeck: Most of the `To[Float|Complex][1|2]D` overloads should be removable now.

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: See my comment above.

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: Yes, that's right. I fiddled a bit with this as well since diags_array shows the following behavior:

If the input is a 1D array, it can be either one diagonal (= offset: int) or a list of constant diagonals (= offset: list[int]). 
But, when the input is 2D, the offset has to be 1D. 

It's possible to enforce this with typing, but this again leads to 2x the amount of overloads.
Besides that, is it the "typers" job to enforce this? Or is this a runtime thing?

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: Damn, I didn't know about the exponential scaling :p As said above, using `onp.ToArray[1|2]D` reduced the number of overloads to 52 already. I'll look into the others as well.

scipy-stubs/sparse/_construct.pyi
jorenham: > But, when the input is 2D, the offset has to be 1D and cannot be of type int.

If that'll double the amount of overloads, then I don't think it's worth it. The potential false negatives aren't that problematic in this case. Generally speaking, users of stubs packages don't mind false negatives at all (but absolutely hate false positives).

That being said, `ToIntStrict1D` might be a bit too strict here: It will reject `npt.NDArray[np.int_]` on `numpy >=2.1, <2.3`, which uses `tuple[int, ...]` as `ndarray` shape-type (which changed to the "gradual" `tuple[Any, ...]` in `numpy >=2.3`). So that's why I was previously using the non-strict `onp.ToInt1D`.

scipy-stubs/sparse/_construct.pyi
jorenham: There's no way around it either, as it is a consequence of how the typing spec defined the overload call evaluation 😅: https://typing.python.org/en/latest/spec/overload.html#overload-call-evaluation

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: I tried this as well, but Pyright doesn't seem to like this. Arrays of type `np.ndarray[tuple[int, int], dtype[...]]` seem to select the wrong overload. It could be that I'm overlooking something...

JulVandenBroeck:
I see that you merged the change to `onp.ToFloat64DType`. Should I update my bramch and remove my custom type in this PR too?

scipy-stubs/sparse/_construct.pyi
```diff
@@ -74,7 +77,9 @@ _FmtDIA: TypeAlias = Literal["dia"]
 _FmtDOK: TypeAlias = Literal["dok"]
 _FmtLIL: TypeAlias = Literal["lil"]
 _FmtNonCOO: TypeAlias = Literal["bsr", "csc", "csr", "dia", "dok", "lil"]
-_FmtNonDIA: TypeAlias = Literal["bsr", "coo", "csc", "csr", "dok", "lil"]
+
+_Diag: TypeAlias = _ToArray1D[_SCT] | _ToArray2D[_SCT]
```
jorenham, line 81: Hmm, I'm not sure about this name; it's not very descriptive. So how about something like `_ToArray1D2D` or ``_ToArrayMax2D` or something?

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +40,16 @@ __all__ = [
     "vstack",
 ]
 
+_Numeric: TypeAlias = npc.number | np.bool_
+_AnyFloat64DTypeNotNone: TypeAlias = type[op.JustFloat] | onp.ToDType[npc.floating64] | f8_code
```
jorenham, line 44: shouldn't be needed anymore after rebasing

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +40,16 @@ __all__ = [
...
 _SCT = TypeVar("_SCT", bound=_Numeric, default=Any)
 _SCT0 = TypeVar("_SCT0", bound=_Numeric)
 _ShapeT = TypeVar("_ShapeT", bound=tuple[int, *tuple[int, ...]], default=tuple[Any, ...])
 
-_Numeric: TypeAlias = npc.number | np.bool_
-
-_ToArray1D: TypeAlias = Seq[_SCT] | onp.CanArrayND[_SCT]
-_ToArray2D: TypeAlias = Seq[Seq[_SCT | int] | onp.CanArrayND[_SCT]] | onp.CanArrayND[_SCT]
+_ToArray1D: TypeAlias = onp.ToArray1D[complex, _SCT]
+_ToArray2D: TypeAlias = onp.ToArray2D[complex, _SCT]
```
jorenham, line 52: This way, `_ToArray1D[np.bool_]` will accept `[1j, 2j]`, is that intentional? Because if not, `onp.CanArray{1,2}D` is probably more appropriate.

JulVandenBroeck:
I think I found a way to differentiate between float-type lists and complex-type Python sequences.
That being said, I get other errors that I can't really explain and are leading me to believe there might be a bug in basedpyright, but I want to check with you to be certain. 
What is the best way to share some code that gives errors?

jorenham:
> I think I found a way to differentiate between float-type and complex-type sequences.

I forgot to mention it earlier, but you can use the `optype.JustFloat` and `optype.JustComplex`  for that ([docs](https://github.com/jorenham/optype#just)). So e.g. `Sequence[op.JustComplex]` accepts `[1j]` but rejects `[1.0]`.

> What is the best way to share some code that gives errors?

If you can make it reproducible without scipy-stubs, then https://basedpyright.com/ is a nice way. Otherwise, you could push it to a new branch on your fork.

> I get other errors that I can't really explain and are leading me to believe there might be a bug in basedpyright, but I want to check with you to be certain.

Coincidentally, I just ran into a very strange case of pyright-specific "spooky action at a distance" myself: https://github.com/scipy/scipy-stubs/pull/741#issuecomment-3059159081
Perhaps they're related 🤷🏻?

JulVandenBroeck:
> I forgot to mention it earlier, but you can use the `optype.JustFloat` and `optype.JustComplex` for that ([docs](https://github.com/jorenham/optype#just)). So e.g. `Sequence[op.JustComplex]` accepts `[1j]` but rejects `[1.0]`.

I want to try to enable the input of mixed arrays like `[[1, 0], [0.5, 1j]]` and I think I found a way by using `np.number[_64Bit, float | _PSCT]`, where `_PSCT` is a type variable bounded to `int | float | complex`. This returns a type conform to `np.float64` if the arrays contain only `int` and `float`, and to `np.complex128` if they contain a mix of `int`, `float`, and `complex`. I'm at the limit of my knowledge here, so it might be too hacky.

> If you can make it reproducible without scipy-stubs, then https://basedpyright.com/ is a nice way. Otherwise, you could push it to a new branch on your fork.

I'll try that tomorrow :)

> Coincidentally, I just ran into a very strange case of pyright-specific "spooky action at a distance" myself: [#741 (comment)](https://github.com/scipy/scipy-stubs/pull/741#issuecomment-3059159081) Perhaps they're related 🤷🏻?

Seems different, since it's happening in the same file as the tests are in, but you never know...

JulVandenBroeck:
Another issue: `mypy` complains about `_ToArray1D2D` requiring two type aliases (first `_PSCT`, then `_SCT`), while `pyright` accepts only `_SCT` or `[_ToArray1D2D[_SCT, _PSCT]`, so the other way around... Weird stuff going on with `pyright`.

jorenham:
I'll have another look at this tomorrow, sorry for the delay

JulVandenBroeck:
No worries! This is all voluntary, so no rush at all 😄 
I'm just clicking the buttons (request review, in this case) that I think are right for the situation haha

JulVandenBroeck:
Sooo, `uv run basedpyright .` and `uv run basedpyright tests` work now with optype 0.12.0, however, `uv run basedpyright tests/sparse/_coonstruct.pyi`now fails, causing the pre-commit hook to fail as well. I had to commit with `git commit -n ...`. 
Any ideas why this happens now?

jorenham:
> Sooo, `uv run basedpyright .` and `uv run basedpyright tests` work now with optype 0.12.0, however, `uv run basedpyright tests/sparse/_construct.pyi`now fails, causing the pre-commit hook to fail as well. I had to commit with `git commit -n ...`. Any ideas why this happens now?

The bug seems to be related to `TypeVar`s with a `default`, the analysis order of modules and individual overloads, and analysis time. For example in https://github.com/scipy/scipy-stubs/pull/744, even with optype 0.12, the bug was still being triggered. I managed to work around it by moving the overloads around a bit (even though they were disjoint, so it shouldn't affect semantic analysis). It might also help to remove some typevar defaults if possible.

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +38,17 @@ __all__ = [
...
 _SCT0 = TypeVar("_SCT0", bound=_Numeric)
 _ShapeT = TypeVar("_ShapeT", bound=tuple[int, *tuple[int, ...]], default=tuple[Any, ...])
 
-_Numeric: TypeAlias = npc.number | np.bool_
-
 _ToArray1D: TypeAlias = Seq[_SCT] | onp.CanArrayND[_SCT]
-_ToArray2D: TypeAlias = Seq[Seq[_SCT | int] | onp.CanArrayND[_SCT]] | onp.CanArrayND[_SCT]
+_ToArray2D: TypeAlias = Seq[Seq[_SCT] | onp.CanArray1D[_SCT]] | onp.CanArray2D[_SCT]
+_ToArray1D2D: TypeAlias = _ToArray1D[_SCT] | _ToArray2D[_SCT]
+_CanArray1D2D: TypeAlias = onp.CanArray1D[_SCT] | onp.CanArray2D[_SCT]
```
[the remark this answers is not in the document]
JulVandenBroeck: Nice! That's definitely better.

scipy-stubs/sparse/_construct.pyi
```diff
@@ -85,265 +89,856 @@ class _DataSampler(Protocol):
...
-    offsets: onp.ToInt | onp.ToInt1D = 0,
+    offsets: _Offsets = 0,
     shape: _ToShape2D | None = None,
-    format: _FmtNonDIA,
+    format: _FmtDIA | None = None,
     dtype: None = None,
-) -> _SpArray2D[_SCT]: ...
-@overload  # diagonals: <unknown>, format: {"dia", None} = ..., dtype: int
+) -> _DIAArray[_SCT]: ...
+@overload  # diagonals: <known>, format: {"dia", None} = ..., dtype: bool-like
```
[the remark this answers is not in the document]
JulVandenBroeck: I agree that the comments are not up-to-date anymore, but I like having a description for the each relevant argument. I updated the comments in f643769; are they OK like that?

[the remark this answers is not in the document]
JulVandenBroeck: (I reverted back to `<known>`, instead of `<known dtype>` to stick with the original comments.)

jorenham (commented):
nitpick: the overload ordering of `eye` and `eye_array` don't really match right now

jorenham:
> I think we're nearing the end of this PR 😅

I think so too

JulVandenBroeck:
```
 Error: No overloads for "diags" match the provided arguments (reportCallIssue)
/home/runner/work/scipy-stubs/scipy-stubs/tests/sparse/test_construct.pyi:67:13 - error: "assert_type" mismatch: expected "bsr_matrix[floating[_32Bit]]" but received "Unknown" (reportAssertTypeFailure)
Error: "assert_type" mismatch: expected "bsr_matrix[floating[_32Bit]]" but received "Unknown" (reportAssertTypeFailure)
/home/runner/work/scipy-stubs/scipy-stubs/tests/sparse/test_construct.pyi:67:43 - error: Argument of type "Literal['bsr']" cannot be assigned to parameter "format" of type "_FmtDIA | None" in function "diags"
  Type "Literal['bsr']" is not assignable to type "_FmtDIA | None"
    "Literal['bsr']" is not assignable to "None"
    "Literal['bsr']" is not assignable to type "_FmtDIA" (reportArgumentType)
```

I have no clue why this error is now triggered... Why is overload 2 not checked, which has `format: _FmtBSR`? I'm going to bed... 😅

JulVandenBroeck:
That being said, I restructured all functions in `_construct.pyi` so that overloads are bundled in groups in which all argument types are the same, except for the format. This is more visually pleasing, and also easier to wrap my mind around. I slightly tweaked some overloads as well, since I found an error in the stubs for `random` and `rand` (`random_state` is not an argument anymore).

jorenham:
> That being said, I restructured all functions in _construct.pyi so that overloads are bundled in groups in which all argument types are the same, except for the format. This is more visually pleasing, and also easier to wrap my mind around. I slightly tweaked some overloads as well, since I found an error in the stubs for random and rand (random_state is not an argument anymore).

So the pyright false positives weren't there before the big shuffle?

> I'll have to work on the other comments later, since I'm going on vacation!

Ah good to know; I'll go ahead and release 1.16.0.3 then. Have fun!

scipy-stubs/sparse/_construct.pyi
```diff
@@ -85,265 +90,796 @@ class _DataSampler(Protocol):
...
     /,
     *,
-    offsets: onp.ToInt | onp.ToInt1D = 0,
+    offsets: _Offsets = 0,
     shape: _ToShape2D | None = None,
     format: _FmtDIA | None = None,
     dtype: None = None,
-) -> dia_array[_SCT]: ...
-@overload  # diagonals: <known>, dtype: None = ..., format: <otherwise>
+) -> _DIAArray[np.float64] | _DIAArray[np.complex128]: ...
```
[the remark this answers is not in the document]
JulVandenBroeck: Hmm, in my work (doing a PhD in computational quantum physics) I almost exclusively use complex sparse arrays; then again, I only use NumPy arrays for initialization, never Python lists. 
So, it is up to you in my opinion. I kinda agree that the union is ugly...

jorenham:
> (`random_state` is not an argument anymore).

The `@_transition_to_rng` decorator adds it :)

JulVandenBroeck:
`So the pyright false positives weren't there before the big shuffle?`

Nope... But they seem to be very sensitive to slight tweaks.

`Ah good to know; I'll go ahead and release 1.16.0.3 then. Have fun!`

Thanks! I kinda figured you were waiting for this PR to release the next version haha, sorry for the delay. On the other hand, I still want to fix the other functions in `_construct.pyi` as well, since I'm getting errors with "block_diag" in my code, where I use specific formats. It will be nice to combine all changes into a new release then.

jorenham:
> Nope... But they seem to be very sensitive to slight tweaks.

It must be a quantum bug: I've seen it causing spooky action at a distance before, and this sounds like a case of decoherence. So putting it in the (dilution) fridge sounds like a good idea to me 👌🏻.

JulVandenBroeck:
Turns out I'm just an idiot (accidentally removed some overloads during refactoring + added some `*` and `/` tokens that were out of place, because I learned about them approx. last week for the first time)! Happy ending :p
The additional unit tests are for later...

JulVandenBroeck:
You can! I don't know how pressing these unit tests are. Maybe it can be marked as a TODO with an issue?

scipy-stubs/sparse/_construct.pyi
```diff
@@ -38,15 +38,17 @@ __all__ = [
     "vstack",
 ]
 
+_Numeric: TypeAlias = np.bool_ | npc.number
+
 _T = TypeVar("_T")
 _SCT = TypeVar("_SCT", bound=_Numeric, default=Any)
 _SCT0 = TypeVar("_SCT0", bound=_Numeric)
 _ShapeT = TypeVar("_ShapeT", bound=tuple[int, *tuple[int, ...]], default=tuple[Any, ...])
```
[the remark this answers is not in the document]
jorenham: In `.pyi` that's not a problem; forward references are fine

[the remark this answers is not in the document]
jorenham: I kinda see how it can be cleaner to keep the typevars and type aliases separated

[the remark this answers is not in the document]
jorenham: feel free to ignore it though

JulVandenBroeck:
I added TODOs and moved `_Numeric` to the other `TypeAlias`es. I'm leaving now, so if all tests pass you can merge 😄

PR #741, opened 2025-07-10
✨ `linalg`: improved basic `solve*` annotations

jorenham:
This improves the following "basic" `scipy.linalg` functions:

- `solve`
- `solve_triangular`
- `solve_banded`
- `solveh_banded`
- `solve_toeplitz`
- `solve_circulant`

A total of 240 type-tests have been added for these functions.

PR #742, opened 2025-07-13
🤡 workaround for path-dependent pyright bug by replicating `CanArrayND`

jorenham:
This works around a Pyright bug where it reports false positives if run as e.g. `pyright tests`, instead of just `pyright .`. I believe this is the same bug as the one in that from #741. 

The bug seems to have to do with type parameter defaults: It has to do with generic protocols that have generic type arguments, that sometimes (depending on the order of analysis) falsely reject assignment to it.
This is the case for `optype.nunpy.CanArray`, which has two type parameters with defaults. When running `pyright tests`, Pyright sometimes incorrectly does not allow assigning compatible `np.ndarray` instances to this `CanArray` protocol. I've only seen this occur in overloads, but I'm not sure if that's relevant here.

Either way, I worked around this by replacing the `optype.numpy.CanArray` with a custom `scipy._typing.CanArray` protocol, that's almost equivelent, apart from its first type parameter not having a default. I.e. it has two type parameters, and only one of them has a default.

Note that currently, even without this PR, `pyright .` works fine, and so does `pyright scipy-stubs`, and `pyright scipy-stubs tests`. But without this workaround, `pyright tests` and `pyright tests scipy-stubs` will report false positive errors.

I'm not sure if I'm going to merge this, as I'm not sure whether the mess this causes is worth it.


JulVandenBroeck:
I looked into this PR and I notice the same difference in behavior for `pyright .` vs. `pyright tests` in #733; no problems in the first case...
I'll have to wait until this PR is merged, I guess :)

PR #744, opened 2025-07-14
✨ `linalg`: improved `det` annotations

jorenham:


jorenham:
Since this appears to trigger the path-dependent pyright bug, this'll have to wait until the next optype release, as explained in #745.

jorenham:
still that pyright issue 😞

PR #769, opened 2025-07-22
💚 Fix ignored NumPy testing matrix

jorenham:
Hopefully this also fixed the pyright's CI failures in #768.

---

This will have to wait until optype 0.13 is released, which fundamentally fixes the numpy compatibility issues through the new [`numpy_typing_compat`](https://github.com/jorenham/numpy-typing-compat) package.


jorenham:
Yikes, there apparently were quite fundamental errors on older numpy versions that were previously hidden... I believe it has to do with the numpy version constants in `optype`, that aren't properly inferred by static type-checkers.

The only way I can think of that could reliably fix this problem, is by creating a new library, e.g. `numpy-typing-compat`, that provides those constants, and release several versions of it that are directly tied to (minor) numpy versions. That way, optype would use `numpy-typing-compat` as a dependency, and dependency managers will install the version of `numpy-typing-compat` that matches the `numpy` version installed in the environment. 

I tried solving this by using build scripts before, but those aren't ran in the same environment as where numpy is installed, and could also be run *before* numpy is installed. So I have up on that. See jorenham/optype#335 for details.

jorenham:
The remaining `numpy<2.1` test failures are limited to `linalg._basic` and `stats._entropy`. These failures are, however, indicative of *actual* underlying compatibility issues. I believe they're all related to shape-typing, so it's likely that these failures are related to `ndarray`'s invariant shape type prior to 2.1.

jorenham:
We're now at the point that we only have failing tests on `numpy<2.1` and only in `stats/_entropy.pyi`. Hopefully that actually means that we're almost there then.

jorenham:
All green! I suppose I'll publish the new optype release then.

jorenham:
oh and note to self: don't forget to restore `ABCPolyBase`, which now exists in `numpy_typing_compat`

PR #774, opened 2025-07-29
Add `hstack` and `vstack` overloads for different sparse formats

JulVandenBroeck:
Back at it again 😎 

This PR should be straight forward; the only thing I'm not sure of is the removal of `Any` from the type aliases `_SpBase`, `_SpMatrix`, and `_SpArray`. What would be the consequence of this?

Thanks!


jorenham:
> I'm not sure of is the removal of `Any` from the type aliases `_SpBase`, `_SpMatrix`, and `_SpArray`. What would be the consequence of this?

I guess we'll find out 🤷🏻

jorenham:
Thanks! And welcome back!

PR #775, opened 2025-07-29
🏷️ added additional overloads and improved existing stubs of block-related sparse construction functions

JulVandenBroeck:
I added all formats for the final functions in `_construct.pyi`.

Let me know if you spot any errors! 😄

### DIFF
--- a/scipy-stubs/sparse/_construct.pyi
+++ b/scipy-stubs/sparse/_construct.pyi
@@ -3,7 +3,6 @@ from collections.abc import Callable, Iterable, Sequence as Seq
 from typing import Any, Literal, Never, Protocol, TypeAlias, TypeVar, overload, type_check_only
 
 import numpy as np
-import numpy.typing as npt
 import optype.numpy as onp
 import optype.numpy.compat as npc
 import optype.typing as opt


JulVandenBroeck:
Most errors are related to `mypy` not recognizing `[[<sparse array>], [None]]` as type `list[list[<sparse array> | None]]`. Instead, the type is marked as `list[object]`. It does work for `pyright` though.
Simply removing the `None` entries from the test cases is a simple workaround.

There is still an error left that I have to investigate.

JulVandenBroeck:
`mpy` < `pyright` ? 👀

jorenham:
> Most errors are related to `mypy` not recognizing `[[<sparse array>], [None]]` as type `list[list[<sparse array> | None]]`. Instead, the type is marked as `list[object]`. It does work for `pyright` though. Simply removing the `None` entries from the test cases is a simple workaround.
> 
> There is still an error left that I have to investigate.

That's probably a classic case of "join vs union", which the (based)pyright docs explain quite nicely: https://docs.basedpyright.com/v1.18.2/usage/mypy-comparison/#unions-vs-joins

I believe that mypy intends to address this class of issues though: https://github.com/python/mypy/issues/12056.

scipy-stubs/sparse/_construct.pyi
```diff
@@ -1643,117 +1648,371 @@ def vstack(blocks: Seq[_SpArray], format: _FmtLIL, dtype: npt.DTypeLike) -> _LIL
...
+@overload  # mats: <array, unknown>, format: "dok", dtype: <known>
+def block_diag(mats: Iterable[sparray], format: _FmtDOK, dtype: onp.ToDType[_SCT]) -> _DOKArray2D[_SCT]: ...
+@overload  # mats: <array, unknown>, format: "lil", dtype: <known>
+def block_diag(mats: Iterable[sparray], format: _FmtLIL, dtype: onp.ToDType[_SCT]) -> _LILArray[_SCT]: ...
 
 #
 @overload  # mats: <unknown, known>, format: <default>, dtype: None
 def block_diag(
-    mats: Iterable[_spbase[_SCT] | onp.ArrayND[_SCT]], format: _FmtCOO | None = None, dtype: None = None
+    mats: Iterable[_spbase[_SCT] | onp.ToArrayND[_SCT]], format: _FmtCOO | None = None, dtype: None = None
```
jorenham, line 1901: I suppose a type alias could help clean this up a bit

scipy-stubs/sparse/_construct.pyi
```diff
@@ -1643,117 +1648,371 @@ def vstack(blocks: Seq[_SpArray], format: _FmtLIL, dtype: npt.DTypeLike) -> _LIL
...
-def block_diag(mats: Iterable[sparray], format: _FmtCOO | None = None, *, dtype: onp.ToDType[_SCT]) -> _COOArray2D[_SCT]: ...
-@overload  # mats: <array, unknown>, format: <default>, dtype: <known> (positional)
-def block_diag(mats: Iterable[sparray], format: _FmtCOO | None, dtype: onp.ToDType[_SCT]) -> _COOArray2D[_SCT]: ...
-
-#
-@overload  # mats: <matrix, unknown>, format: <default>, dtype: <known> (keyword)
+@overload  # mats: <unknown, unknown>, format: <default>, dtype: <known>
 def block_diag(
-    mats: Iterable[spmatrix | onp.ArrayND[_Numeric] | complex | Seq[onp.ToComplex] | Seq[onp.ToComplex1D]],
+    mats: Iterable[_spbase | onp.ArrayND[_Numeric] | complex | Seq[onp.ToComplex] | Seq[onp.ToComplex1D]],
```
jorenham, line 1931: This one could also use a type alias

jorenham (commented):
Does the `block_diag` return type (matrix vs array) still depend on the input, or did that maybe change in the meantime?

scipy-stubs/sparse/_construct.pyi
```diff
@@ -1636,117 +1645,306 @@ def vstack(blocks: Seq[sparray], format: _FmtLIL, dtype: npt.DTypeLike) -> _LILA
...
+@overload  # blocks: <unknown, unknown>, format: "dia", dtype: bool-like
+def block_array(blocks: _ToBlocksUnkown, *, format: _FmtDIA, dtype: onp.AnyBoolDType) -> _DIAArray[np.bool_]: ...
+@overload  # blocks: <unknown, unknown>, format: "dok", dtype: bool-like
+def block_array(blocks: _ToBlocksUnkown, *, format: _FmtDOK, dtype: onp.AnyBoolDType) -> _DOKArray2D[np.bool_]: ...
+@overload  # blocks: <unknown, unknown>, format: "lil", dtype: bool-like
+def block_array(blocks: _ToBlocksUnkown, *, format: _FmtLIL, dtype: onp.AnyBoolDType) -> _LILArray[np.bool_]: ...
+
+#
+@overload  # blocks: <known, int_>, format: <default>, dtype: int-like
+def block_array(blocks: _ToBlocksCanStackAs[np.int_, _T], *, format: None = None, dtype: onp.AnyIntDType) -> _T: ...
```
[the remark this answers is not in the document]
JulVandenBroeck: Ah, I see. I'm going to leave the overloads like this, since I'm using the same int-types for the other functions. Instead, I've changed the test case. The stubs are then "most correct" for `numpy>=2.2`.

JulVandenBroeck:
> Hmm yea, those commits don't look right. Did you by any chance rebase and then `git push` instead of `git push -f`?
I think you nailed it, but now I'm in too deep apparently 😆.
Will it cause problems with the merge? Otherwise, I'll make a new branch and corresponding PR...

.github/workflows/ci.yml
```diff
@@ -65,8 +65,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4.2.2
 
-      - name: setup uv
-        uses: astral-sh/setup-uv@v6.4.3
+      - uses: astral-sh/setup-uv@v6.4.3
```
jorenham, line 68: seems unrelated

scipy-stubs/sparse/_construct.pyi
```diff
@@ -47,6 +47,7 @@ _Numeric: TypeAlias = npc.number | np.bool_
 
 _SpBase2D: TypeAlias = _spbase[_SCT, tuple[int, int]]
 _SpArray2D: TypeAlias = sparray[_SCT, tuple[int, int]]
+_SpMatrix: TypeAlias = spmatrix[_SCT]
```
jorenham, line 50: this seems a bit redundant

scipy-stubs/sparse/_construct.pyi
```diff
@@ -76,7 +77,15 @@ _Offsets: TypeAlias = onp.ToInt | onp.ToInt1D
 
 _DataRVS: TypeAlias = Callable[[int], onp.ArrayND[_Numeric]]
 
-_ToBlocks: TypeAlias = Seq[Seq[_spbase[_SCT] | None]] | onp.ArrayND[np.object_]
+_ToBlocksArray: TypeAlias = onp.ArrayND[np.object_]
+_ToBlocksSPArray: TypeAlias = Seq[Seq[_SpArray2D[_SCT] | None]]
+_ToBlocksSPMatrix: TypeAlias = Seq[Seq[_SpMatrix[_SCT] | None]]
+_ToBlocksCanStack: TypeAlias = Seq[Seq[_CanStack[_T] | None]]
+_ToBlocksCanStackAs: TypeAlias = Seq[Seq[_CanStackAs[_SCT0, _T] | None]]
```
jorenham, line 84: Maybe we could combine these four aliases:

```pyi
_ToBlocks: TypeAlias = Seq[Seq[_T | None]]
```

Or otherwise use `_ToBlocks` as an intermediary alias to simplify their definitions.

scipy-stubs/sparse/_construct.pyi
```diff
@@ -76,7 +77,15 @@ _Offsets: TypeAlias = onp.ToInt | onp.ToInt1D
...
+_ToBlocksUnkown: TypeAlias = _ToBlocksSPArray | _ToBlocksArray
+
+_ToMatsDiagKnown: TypeAlias = Iterable[_spbase[_SCT] | onp.ToArrayND[_SCT]]
```
jorenham, line 87: I suppose that because the generic type parameter already implies that the type is known, there's no real need for that explicit `*Known` suffix. But it's also fine to leave it like this AFAIK 🤷🏻

scipy-stubs/sparse/_construct.pyi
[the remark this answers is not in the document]
jorenham: Well, we follow [SPEC 0](https://scientific-python.org/specs/spec-0000/), so we support all python versions equally :). 

And to let you in on a little secret: `np.int64` is actually equivalent to `np.int_` (and `np.intp`) (in `numpy>=2.2`). It's just "pretending" that it isn't, but `_32Bit | _64Bit` is actually the same as `_64Bit`, because `_32Bit` is a subtype of `_64Bit` ([source](https://github.com/numpy/numpy/blob/9da497f5c88a81669c954c3a9e328a15ce0a7220/numpy/_typing/_nbit_base.py#L76-L84)). And I think that the reason we've been getting away with it, is because almost everyone has a 64bit system nowadays, so it's only wrong for those that still use a 32bit system. 
There are probably way more 64bit `numpy<2.2` users then there are 32bit users in general, so changing this to `int64` would be the least painful.

scipy-stubs/sparse/_construct.pyi
JulVandenBroeck: Agreed! I'll use it as an intermediary, otherwise it'll still be quite verbose.

scipy-stubs/sparse/_construct.pyi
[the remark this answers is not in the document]
JulVandenBroeck: Ok, I understand! I've changed the type to int64 and altered the test case accordingly.

jorenham:
Hmm, I just compared the timing results of `uv run basedpyright --stats --verbose scipy-stubs/sparse`, and this change increases the analysis time of `scipy-stubs/sparse/_construct.pyi` from `1038ms` to `1316ms` (+27%). Similarly, this increases the analysis time of `tests/sparse/test_construct.pyi` from `847ms` to `964ms` (+14%). There's little difference in outcomes when when repeating the measurements (I've seen +-25ms at most), so the hypothesis that this is causes a performance regression appears to be statistically significant. 

I don't think that the test slowdown is any reason for concern, especially if you consider the tests that were added. But the `+300ms` of `_construct.pyi` itself can be pretty annoying for DX. For me it causes my laptop fan to go into helicopter mode for a couple of seconds each time that I make an edit to `_construct.pyi` with the basedpyright extension enabled, which wasn't the case before. 

For context, these are the top 10 slowest analysis times per file in `scipy-stubs`:

```
1314ms: scipy-stubs/sparse/_construct.pyi
 919ms: scipy-stubs/signal/_signaltools.pyi
 460ms: scipy-stubs/linalg/_basic.pyi
 279ms: scipy-stubs/sparse/_base.pyi
 278ms: scipy-stubs/special/_ufuncs.pyi
 252ms: scipy-stubs/stats/_distribution_infrastructure.pyi
 227ms: scipy-stubs/sparse/linalg/_dsolve/linsolve.pyi
 221ms: scipy-stubs/sparse/_dok.pyi
 197ms: scipy-stubs/stats/_entropy.pyi
 190ms: scipy-stubs/ndimage/_filters.pyi
 190ms: scipy-stubs/special/_logsumexp.pyi
```

So even before this change, `_construct.pyi` was already the bottleneck, and this increases its bottlenecknessitude by +25% or so. 

I'm also worried that this will affect scipy-stubs users that use `scipy.sparse` a lot. These kind of things have the tendency to snowball into additional seconds or minutes of CI runtime, and that can be pretty costly (and very annoying in general).

---

I'll see if I can figure out an easy way to improve the analysis time. Sometimes it's as easy as tweaking a single type alias, or swapping some overloads, so let's hope that's also the case here. In this case, I suspect that it has to do with the overload-overlap detection algorithm, which has an exponential time-complexity in terms of overload count and the amount of protocol members of the input types. So it might help to replace a couple of `onp.CanArrayND` with `onp.CanArray`. But to be honest, optimizing for type-checker performance isn't something I have much experience in, so I might very well be wrong here 🤷🏻.

Anyway, it's just something to keep in mind. 

---

It would be pretty cool to have a script that warns about performance regressions in CI 🤔. I guess I'll open an issue for it.

JulVandenBroeck:
Thanks for the detailed analysis!

> It would be pretty cool to have a script that warns about performance regressions in CI 🤔. I guess I'll open an issue for it.

Indeed; I didn't notice this slowdown, so this warning would be quite nice!

> Since these changes aren't directly related to this PR, I first thought I'd create a separate PR for this. But fix 2. in particular would've caused nasty merge conflicts. And even though it's not very clean, it's fine by me to put these performance tweaks in this PR. Those two should be enough to counteract the +300 ms :)

Sounds good! I'll start implementing these suggestions soon :)

PR #776, opened 2025-07-30
🐛 `interpolate`: Fix incorrect return type for `make_lsq_spline`

elael:
The overloads for `make_lsq_spline` in `_bsplines.pyi` had their return types for float and complex `y` inputs swapped.

This change corrects the type hints to ensure that:
- `make_lsq_spline` with a float `y` array correctly returns `BSpline[np.float64]`.
- `make_lsq_spline` with a complex `y` array correctly returns `BSpline[np.complex128]`.

Hopefully this resolves incorrect type checker errors for valid float-only spline creation. 😄

### DIFF
--- a/scipy-stubs/interpolate/_bsplines.pyi
+++ b/scipy-stubs/interpolate/_bsplines.pyi
@@ -184,7 +184,7 @@ def make_smoothing_spline(
     lam: onp.ToFloat | None = None,
     *,
     axis: op.CanIndex = 0,
-) -> BSpline[np.float64]: ...
+) -> BSpline[np.complex128]: ...
 
 #
 def fpcheck(x: onp.ToFloat1D, t: onp.ToFloat1D, k: onp.ToJustInt) -> None: ...  # undocumented
--- a/scipy-stubs/interpolate/_bsplines.pyi
+++ b/scipy-stubs/interpolate/_bsplines.pyi
@@ -162,7 +162,7 @@ def make_lsq_spline(
     check_finite: onp.ToBool = True,
     *,
     method: _LSQMethod = "qr",
-) -> BSpline[np.complex128]: ...
+) -> BSpline[np.float64]: ...
 @overload
 def make_lsq_spline(
     x: onp.ToFloat1D,
--- a/scipy-stubs/interpolate/_bsplines.pyi
+++ b/scipy-stubs/interpolate/_bsplines.pyi
@@ -142,7 +142,7 @@ def make_interp_spline(
 @overload
 def make_lsq_spline(
     x: onp.ToFloat1D,
-    y: onp.ToJustComplexND,
+    y: onp.ToFloatND,
     t: onp.ToFloat1D,
     k: op.CanIndex = 3,
     w: onp.ToFloat1D | None = None,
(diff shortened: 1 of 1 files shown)


scipy-stubs/interpolate/_bsplines.pyi
```diff
@@ -184,7 +184,7 @@ def make_smoothing_spline(
     lam: onp.ToFloat | None = None,
     *,
     axis: op.CanIndex = 0,
-) -> BSpline[np.float64]: ...
+) -> BSpline[np.complex128]: ...
```
jorenham, line 187: Hmm, I don't think that's right actually. The type argument to `BSpline` applies to the dtype of the coefficient array (`c`) of the returned `BSpline` object. But as the comment above explains this function explains, even if `x` and/or `y` are complex, then this will still return a `BSpline` object with `float64` coefficients, and should therefore be annotated as `BSpline[np.float64]`

```pycon
>>> import numpy as np; from scipy.interpolate import make_smoothing_spline
>>> x = np.arange(10)
>>> y = np.linspace(0, 1, 10) ** 2
>>> make_smoothing_spline(x, y).c.dtype
dtype('float64')
>>> make_smoothing_spline(x + 1j, y + 1j).c.dtype
/home/joren/.pyenv/versions/3.13.5/lib/python3.13/site-packages/scipy/interpolate/_bsplines.py:2282: ComplexWarning: Casting complex values to real discards the imaginary part
  x = np.ascontiguousarray(x, dtype=float)
/home/joren/.pyenv/versions/3.13.5/lib/python3.13/site-packages/scipy/interpolate/_bsplines.py:2283: ComplexWarning: Casting complex values to real discards the imaginary part
  y = np.ascontiguousarray(y, dtype=float)
dtype('float64')
```

This seems to be the case for both scipy `1.16.0` and `1.16.1`.

scipy-stubs/interpolate/_bsplines.pyi
```diff
@@ -162,7 +162,7 @@ def make_lsq_spline(
     check_finite: onp.ToBool = True,
     *,
     method: _LSQMethod = "qr",
-) -> BSpline[np.complex128]: ...
+) -> BSpline[np.float64]: ...
```
jorenham, line 165: Yea you're right; if `y` is real, then the spline coefficients will also be real. However, that still leaves the overload above this one, which will *only* be used for complex `y`. In that, the returned spline coefs will be complex too. However, it is currently incorrectly annotated to return `BSpline[np.float64]` instead of `BSpline[np.complex128]`.

So instead of changing this return value to `float64`, it's probably easier to swap the annotation of `y` with the overload above. That way the float overload comes first, and the complex one second, which feels a bit more natural to me.

jorenham (changes requested):
Thanks for this; good catch! I see that `make_lsq_spline` is indeed incorrectly annotated here. But I don't think that `make_smoothing_spline` should be changed.

scipy-stubs/interpolate/_bsplines.pyi
elael: Thanks for the keen eye. I intended to modify the other `make_lsq_spline` overload.

jorenham: Ah yea that makes sense. There's too many damn overloads to keep track of, so it's certainly an easy mistake to make haha

jorenham:
Thanks Eduardo, and welcome aboard :)

PR #836, opened 2025-08-24
👷 mypy_primer: custom project selection

jorenham:


jorenham:
Maybe we should skip colour, or at least when using mypy, cus it's by far the slowest one (almost 10 min) 🤔

jorenham:
Hmm I guess I didn't break it enough

jorenham:
Ah; this of course won't work properly if many of the projects ignore unresolved imports...

PR #837, opened 2025-08-24
👷 mypy_primer: sharding

jorenham:
let's see if this will speed things up


jorenham:
with 2 shards it runs in 8m 5s, let's try 3 shards now

jorenham:
2 and 3 shards does make make a difference apparently, so let's stick with 2 shards

PR #839, opened 2025-08-24
👷 mypy_primer: simplified workflow logic

jorenham:
It now uses a single shard, and the `colour` project was removed (for now) because it was taking up the vast majority of the runtime.


jorenham:
ah of course; this has to be merged first for the comment workflow to be able to do it's thing

PR #840, opened 2025-08-24
👷 mypy_primer: use pyright instead of mypy

jorenham:
mostly because it's (supposed to be) faster


jorenham:
10 minutes?! yikes...

let's see it's still slower than mypy with 1.1.403 (1.1.404 has a performance regression)

jorenham:
well, that didn't help..

PR #860, opened 2025-09-05
:sparkles: `optimize`: support plural bounds in `linprog`

Rupt:
- Annotate the 'sequence of bound' option to [`scipy.optimize.linprog`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html), which is documented as:

> **bounds: sequence, optional**
>
> > A sequence of (min, max) pairs for each element in x, defining the minimum and maximum values of that decision variable. If a single tuple (min, max) is provided, then min and max will serve as bounds for all decision variables.

We cannot use the `Bounds` annotation from `scipy-stubs/optimize/_typing.pyi`, because it is a union with `scipy.optimize.Bounds`, which `linprog` does not support. For example:

```console
$ uv run python -c "import scipy; scipy.optimize.linprog(range(2), bounds=(None, None))"
$ uv run python -c "import scipy; scipy.optimize.linprog(range(2), bounds=[(None, None), (None, None)])"
$ uv run python -c "import scipy; scipy.optimize.linprog(range(2), bounds=scipy.optimize.Bounds())"
Traceback (most recent call last):
  File "/home/rupert/scipy-stubs/.venv/lib/python3.11/site-packages/scipy/optimize/_linprog_util.py", line 431, in _clean_inputs
    bounds_conv = np.atleast_2d(np.array(bounds, dtype=float))
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: float() argument must be a string or a real number, not 'Bounds'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/rupert/scipy-stubs/.venv/lib/python3.11/site-packages/scipy/optimize/_linprog.py", line 649, in linprog
    lp, solver_options = _parse_linprog(lp, options, meth)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rupert/scipy-stubs/.venv/lib/python3.11/site-packages/scipy/optimize/_linprog_util.py", line 1026, in _parse_linprog
    lp = _clean_inputs(lp._replace(A_ub=A_ub, A_eq=A_eq))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rupert/scipy-stubs/.venv/lib/python3.11/site-packages/scipy/optimize/_linprog_util.py", line 437, in _clean_inputs
    raise TypeError(
TypeError: Invalid input for linprog: unable to interpret bounds, check values and dimensions: float() argument must be a string or a real number, not 'Bounds'
```

Thank you for maintaining scipy-stubs!


jorenham (approved):
Thanks for this!

These tests made me realize that `linprog` isn't returning the `optimize._linprog._OptimizeResult` specialization 🤔.

But that's probably something we should address in a separate PR; I'll open an issue for it.

PR #916, opened 2025-10-04
🐛 `io`: Fix wrong return type on `scipy.io.wavfile`

mochaaP:
Trivial change to match `scipy.io.wavfile.read` to the actual API.

https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html


jorenham:
Good catch, thanks @mochaaP!

PR #956, opened 2025-10-24
💡 Update pyrefly suppressions to reflect rename of `typed-dict-key-error` to `bad-typed-dict-key`

rchen152:
I renamed this error kind in pyrefly 0.38.1. (We're temporarily allowing suppressions with both names.)


jorenham:
Much appreciated!

And the name is indeed better this way.

Issue #938, opened 2025-10-14
Add a readme section explaining why scipy-stubs isn't part of scipy


jorenham:
There are several (good) reasons, so it's probably too big to put in the FAQ. It's also a pretty common question, and the answer isn't obvious.

This section could be based on https://github.com/scipy/scipy-stubs/discussions/884.

PR #977, opened 2025-11-12
📝 Readme: Document reasons for scipy-stubs as a separate package

Abhinavexists:
as per the suggestion of #938
Added section explaining the reasons for maintaining scipy-stubs as a separate package

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ conda install conda-forge::scipy-typed
 
 ## Why is scipy-stubs a separate package?
 
-These are only a subset of the many other important reasons why `scipy-stubs` is maintained as a separate package rather than being integrated into SciPy itself.
+Here are some of the key reasons why `scipy-stubs` is maintained as a separate package rather than being integrated into SciPy itself.
 
 ### Separation of concerns
 


README.md
```diff
@@ -92,6 +92,26 @@ It's also possible to install both `scipy` and `scipy-stubs` together through th
 conda install conda-forge::scipy-typed
 ```
 
+## Why is scipy-stubs a separate package?
+
+There are several important reasons why `scipy-stubs` is maintained as a separate package rather than being integrated into SciPy itself.
+
+### Separation of concerns
+
+If `scipy-stubs` were part of SciPy, every contributor would need to be familiar with type stubs. Writing high-quality stubs is complex and requires significant typing expertise ([example](link-to-example)). Requiring all SciPy developers to become typing experts would likely reduce overall willingness to contribute. By keeping stubs separate, SciPy contributors can focus on scientific functionality while typing specialists handle the stubs, with each group reviewing changes in their domain of expertise.
```
jorenham: > Writing high-quality stubs is complex and requires significant typing expertise

True as this is, I'm a bit worried that this formulation might be demotivating for bypassers considering to contribute.  So how about something like "Some parts of scipy-stubs are very complex and maintaining it requires a deep understanding of the typing intricacies involved."?

README.md
```diff
@@ -92,6 +92,26 @@ It's also possible to install both `scipy` and `scipy-stubs` together through th
...
+
+A significant portion of SciPy users don't use type checking. Without `scipy-stubs` installed, type checkers won't analyze scipy-related code. Some code patterns may be technically type-unsafe but work perfectly fine in practice installing `scipy-stubs` would flag these cases. By keeping stubs separate, users retain full control over whether to enable type checking for their SciPy code.
+
+### Independent release cycles
+
+The `scipy-stubs` versioning scheme uses four numbers: `{scipy_version}.{stubs_version}`. The first three match the SciPy semver version, while the fourth is specific to stub releases. Since stub releases occur more frequently than SciPy releases, this independence allows typing improvements and bug fixes to be delivered without waiting for the next SciPy version.
+
+### CI efficiency
+
+The SciPy CI pipeline is comprehensive and takes over 30 minutes to run. The `scipy-stubs` CI is lightweight and completes in under 4 minutes. This difference has a substantial impact on development workflow, enabling rapid iteration and faster delivery of improvements.
```
jorenham: It could ironically be that *I'm* the one hallucinating, but to me, the term "comprehensive" has become a signal that it's written by an LLM, and have grown to be a bit allergic to it because of that. So as unreasonable as it is, it wouldn't be unlikely for other people also have this association.

jorenham: It might help to note the reason that scipy-stubs' CI is that much faster, i.e. that no compilation is required, since we can just use SciPy's pre-built wheels.

Abhinavexists:
> Thanks for this; much appreciated!
> 
> Overall it looks good; and I like your choice of subsections.
> 
> One global nitpick that could help for reviewing and DX, is to limit the max line lentgth a bit. It's not something we actively enforce, but so far, I've used to convention of adding a linebreak after each sentence. In case of very long sentences, I sometimes also break after a comma.
> 
> Recently another important reason for having a separate stubs-only package came up in a discussion on a NumPy pr between me and a very active maintainer of pandas-stubs: [numpy/numpy#30174 (comment)](https://github.com/numpy/numpy/pull/30174#issuecomment-3506664949) It is related to the "independent release cycles" subsection, but perhaps it deserves a separate subsection of its own. What do you think?

Sure, i can work on adding a line break after each sentence so to limit the max line length

https://github.com/numpy/numpy/pull/30174#issuecomment-3506664949
i hope that i understood the reasoning correctly here between you and Irv is to avoid a support policy overhead as they dont support old version stubs , and each stubs are maintained with its current version (current version of numpy with its current numpy-stubs)

Its more of a resource constraint that comes with managing multiple version

i feel like its not something directly required here as a part of this section but as this holds an importance we can add it as a maintainer's note, what are your opinions on this ?

README.md
Abhinavexists: Sure , i am thinking of re-writing it like this 

Some areas of scipy-stubs get pretty intricate, so keeping them in good shape goes more smoothly when you’re comfortable with how the typing pieces interact.

README.md
Abhinavexists: > It could ironically be that _I'm_ the one hallucinating, but to me, the term "comprehensive" has become a signal that it's written by an LLM, and have grown to be a bit allergic to it because of that. So as unreasonable as it is, it wouldn't be unlikely for other people also have this association.

you are right, as few of the correction are made my llm, it can happen to have these repeatatives as this is a readme section and should not have any grammatical errors and i am not that good grammatically,  but sure i will add a alternative.

README.md
jorenham: Yea that's right.

There's for example also [`optype`](https://github.com/jorenham/optype) that scipy-stubs heavily relies on. And although it's all nicely documented (pats self on shoulder), it's a pretty large library, so it takes some time to become familiar with it.

Then there are also a couple of functions and methods with 50+ overloads, and I just don't think it'd be fair to ask scipy contributors to keep those in sync with the runtime, as that would likely take the majority of their time.

README.md
Abhinavexists: totally understandable

Abhinavexists:
@jorenham i have made the fixes

README.md
```diff
@@ -92,6 +92,42 @@ It's also possible to install both `scipy` and `scipy-stubs` together through th
...
+
+The SciPy CI pipeline is large and takes over 30 minutes to run.
+The `scipy-stubs` CI is lightweight and finishes in under 4 minutes.
+This difference has a real impact on development workflow, enabling quicker iteration and faster delivery of changes.
+The speed gap exists mainly because `scipy-stubs` doesn’t require compilation - it can rely on SciPy’s pre-built wheels.
+
+> **Maintainer's note:** A very important point is that scipy-stubs only tracks the current SciPy release and doesn't maintain older versions.
+> SciPy has long support windows and can't change typing behavior quickly. scipy-stubs can.
+> If a typing trick breaks or needs adjustment, a new stub release can go out the same day.
+> That speed removes the overhead of supporting multiple stub versions and keeps the typing layer flexible without adding extra burden to SciPy.
```
jorenham: Hmm, now that I see it, I think this is probably out of scope, after all. Sorry about that.

Abhinavexists:
@jorenham remove the maintainer,s note

jorenham:
Looks good! Let's see what the robots thinks before merging

jorenham:
Lovely; thanks @Abhinavexists !

Issue #991, opened 2025-11-21
DisjointSet.__getitem__ has incorrect type


lispandfound:
The `DisjointSet.__getitem__` method incorrectly sets the return type to `int` when it should probably be the generic `_T`.

The definition is here, 

https://github.com/scipy/scipy-stubs/blob/bb6b6bdaba14eb4f31069eb80407e675d1bef308/scipy-stubs/_lib/_disjoint_set.pyi#L16

Note that the [current implementation](
https://github.com/scipy/scipy/blob/94c3a8076d1756f37abe4ac3c4c7ef7f51eb4616/scipy/_lib/_disjoint_set.py#L116-L137) of `__getitem__` method indicates it is generic in its return type (and this is also my experience using this method).

```python
    def __getitem__(self, x):
        """Find the root element of `x`.

        Parameters
        ----------
        x : hashable object
            Input element.

        Returns
        -------
        root : hashable object
            Root element of `x`.
        """
        if x not in self._indices:
            raise KeyError(x)

        # find by "path halving"
        parents = self._parents
        while self._indices[x] != self._indices[parents[x]]:
            parents[x] = parents[parents[x]]
            x = parents[x]
        return x
```

PR #992, opened 2025-11-21
:bug: `cluster.hierarchy`: use generic return type for `DisjointSet.__getitem__`

lispandfound:
Addresses #991

### DIFF
--- a/tests/cluster/test_hierarchy.pyi
+++ b/tests/cluster/test_hierarchy.pyi
@@ -35,7 +35,6 @@ assert_type(len(disjoint_set_str), int)
 
 # __contains__ accepts an element of the universal set and returns a boolean
 assert_type("a" in disjoint_set_str, bool)
-_ = 1 in disjoint_set_str  # type: ignore[operator]  # pyright: ignore[reportOperatorIssue]
 assert_type(np.int64(2) in disjoint_set_i64, bool)
 
 # __getitem__ returns an element of the universal set


lispandfound:
Yes I will do this, and I will also address the incorrect type in `__init__` you noted in the issue.

lispandfound:
Ok I think this is now in a reviewable state again.

scipy-stubs/_lib/_disjoint_set.pyi
```diff
@@ -1,19 +1,20 @@
...
 class DisjointSet(Generic[_T]):
     n_subsets: int
 
-    def __init__(self, /, elements: _T | None = None) -> None: ...
+    def __init__(self, /, elements: Iterable[_T] | None = None) -> None: ...
     def __iter__(self, /) -> Iterator[_T]: ...
     def __len__(self, /) -> int: ...
-    def __contains__(self, x: object, /) -> bool: ...
-    def __getitem__(self, x: _T, /) -> int: ...
+    def __contains__(self, x: _T, /) -> bool: ...
```
jorenham: This should stay as `x: object`, because `object() in DisjointSet(["some", "stuff"])` works fine.

lispandfound: So I did think this too, and was about to write a comment earlier to that effect. But it isn't how collections in e.g. stdlib typically work. Consider the following,

```python
x: set[str] = {1, 2, 3}

assert 'a' in x
```

This will not typecheck with `mypy --strict` (but not regular `mypy`)

```
error: Non-overlapping container check (element type: "str", container item type: "int")  [comparison-overlap]
Found 1 error in 1 file (checked 1 source file)
```
Even though this does not produce a runtime problem evidently there is a use case for this.

jorenham: > So I did think this too, and was about to write a comment earlier to that effect. But it isn't how collections in e.g. stdlib typically work. Consider the following,
> 
> ```python
> x: set[int] = {1, 2, 3}
> 
> assert 'a' in x
> ```
> 
> This will not typecheck with `mypy --strict` (but not regular `mypy`)
> 
> ```
> error: Non-overlapping container check (element type: "str", container item type: "int")  [comparison-overlap]
> Found 1 error in 1 file (checked 1 source file)
> ```
> Even though this does not produce a runtime problem evidently there is a use case for this.

Ah I see. But the way I see it, that's more of a mypy quirk than a true restriction. I think we should try to model the runtime behavior as closely as possible.

jorenham:
This is looking very good; thanks Jake!

PR #1200, opened 2026-01-19
✨✅ `stats`: improve and test `lmoment`

jorenham:
Because the best function in all of SciPy deserves elaborate stubs ;)

towards #1099


jorenham:
I like the PR number; it's very appropriate if you think about it

PR #1250, opened 2026-01-31
✨ `stats`: made resampling results dimensionally aware (monte_carlo, bootstrap, permutation)

Deshan-5:
While adding stubs for monte_carlo_test, 
I identified the lack of dimensional accuracy in the resampling result types.

This PR introduces a second TypeVar (_DistT) to MonteCarloTestResult, BootstrapResult, and PermutationTestResult.

PR #1251, opened 2026-01-31
`stats`: vectorized `goodness_of_fit` and upgrade to dataclass results.

Deshan-5:
I’ve updated the stubs for goodness_of_fit and fit in scipy.stats to better handle vectorized data.
The main change  i did was moving GoodnessofFitResult from a NamedTuple to a Generic dataclass. 
This matches and allows for much better type-checking when users pass in multi-dimensional arrays.


Deshan-5:
Hi @jorenham,

I am closing this PR to reduce noise. I attempted to add the requested tests here, but I ran into persistent Git synchronization issues on this branch that are causing the CI to fail on errors I've already resolved locally.

To keep things clean, I will close this and focus on adding the tests directly to PR #1250 (or a fresh, clean branch) instead.

Thanks for your patience!

PR #1252, opened 2026-01-31
`stats`: vectorized `ConfidenceInterval` and `BootstrapResult`.

Deshan-5:
I have updated ConfidenceInterval in _common.pyi and BootstrapResult in _resampling.pyi to support vectorized inputs.

Key Changes
Generic ConfidenceInterval: Switched from NamedTuple to a Generic dataclass. 
It now uses a TypeVar to handle both scalars and NumPy arrays.

BootstrapResult Update: Connected the types so that confidence_interval correctly inherits the dimensionality (_FloatNDT) of the bootstrap results.

Used frozen=True on the dataclass to keep the existing read only behavior.

### DIFF
--- a/tests/stats/test_resampling.py
+++ b/tests/stats/test_resampling.py
@@ -1,0 +1,64 @@
+# ruff: noqa: INP001
+from __future__ import annotations
+
+from typing import Any, assert_type
+
+import numpy as np
+import numpy.typing as npt
+import optype.numpy as onp
+
+from scipy import stats
+
+# --- CI Compatibility Helpers ---
+# We use explicit helper functions to provide strict input signatures.
+# This ensures that the CI type-checker resolves arguments as 'NDArray'
+# rather than 'Unknown', preventing 'Any' return type inference failures.
+
+
+def strict_mean(x: npt.NDArray[np.float64]) -> float:
+    """Provides a strict (NDArray) -> float signature for scalar validation."""
+    return float(np.mean(x))
+
+
+def batch_mean(x: npt.NDArray[np.float64], axis: int = -1) -> npt.NDArray[np.float64]:
+    """Provides a vectorized (NDArray, axis) -> NDArray signature."""
+    # We use a type ignore here as np.mean return types vary by NumPy version
+    return np.mean(x, axis=axis)  # type: ignore[return-value]
+
+
+def strict_rvs(
+    size: int | tuple[int, ...] | None = None,
+    *,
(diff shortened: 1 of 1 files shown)


scipy-stubs/stats/_fit.pyi
```diff
@@ -16,6 +17,10 @@ _GOFStatFunc: TypeAlias = Callable[[rv_continuous_frozen, onp.ArrayND[np.float64
 _FitMethod: TypeAlias = Literal["mle", "mse"]
 _PlotType: TypeAlias = Literal["hist", "qq", "pp", "cdf"]
 
+# Define fresh local TypeVars to ensure they are "Free" for the Generic class
```
jorenham:
```suggestion
```

scipy-stubs/stats/_fit.pyi
```diff
@@ -28,12 +33,11 @@ class _PXF2n(Protocol):
     def __call__(self, x: onp.ToFloat, arg0: onp.ToFloat, arg1: onp.ToFloat, /, *args: onp.ToFloat) -> np.float64: ...
 
 _PXFT_co = TypeVar("_PXFT_co", bound=Callable[Concatenate[onp.ToFloat, ...], np.float64], default=_PXF1n | _PXF2n, covariant=True)
-_AxesT = TypeVar("_AxesT", default=Any)  # represents a `matplotlib.lines.Axes`
+_AxesT = TypeVar("_AxesT", default=Any)
 
 ###
 
 class FitResult(Generic[_PXFT_co]):
-    # tuple of at least size 1
```
jorenham: revert please

scipy-stubs/stats/_fit.pyi
```diff
@@ -48,21 +52,20 @@ class FitResult(Generic[_PXFT_co]):
...
+    fit_result: FitResult[_PXF2n]
+    statistic: _ST
+    pvalue: _ST
+    null_distribution: _DT
 
 @overload
 def fit(
     dist: rv_discrete,
-    data: onp.ToFloat1D,
+    data: onp.Array1D[np.float64],  # Strict 1D Array
```
jorenham: Are you sure that array-likes such as `[1,2,3,4]` or things with only an `__array__` method can't be used here?

jorenham (changes requested):
Thanks, but as you can see, CI isn't happy. I've also left a couple of comments. Please refrain from making unnecessary changes, as that complicates the reviewing process.

PR #1260, opened 2026-02-05
`stats`: Overload `binned_statistic_2d` and `binned_statistic_dd` so that `expand_binnumbers=True` returns the shape 

### LINKED ISSUE #1259: `stats`:  incorrect return type for `binned_statistic_2d` and `binned_statistic_dd` when `expand_binnumbers=True`
From the [documentation](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.binned_statistic_2d.html#binned-statistic-2d) on `binned_statistic2d`, setting `expand_binnumbers` to `True` will make it so the `binnumber` field in the result is a 2D array of shape `(2, N)` instead of a 1D array of shape `(N,)`.

Currently all invocations of `binned_statistic2d` return a 1D array. We should add an overload to fix this.

`binned_statistic_dd` also has the same issue.

Not sure if this falls under the other issue #1141, but that doesn't reference `expand_binnumbers`.

I have made a PR #1260 to close this.

pavyamsiri:
This should close #1259.

I have also included tests for the overloads.

Not entirely satisfied with the names of the named tuple results so happy to change it if there are any better suggestions.

### DIFF
--- a/scipy-stubs/stats/_binned_statistic.pyi
+++ b/scipy-stubs/stats/_binned_statistic.pyi
@@ -1,5 +1,6 @@
 from collections.abc import Callable, Sequence
-from typing import Any, Generic, Literal, NamedTuple, TypeAlias, TypeVar, overload
+from typing import Generic, Literal, NamedTuple, TypeAlias, overload
+from typing_extensions import TypeVar
 
 import numpy as np
 import optype.numpy as onp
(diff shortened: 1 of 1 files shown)


pavyamsiri:
Ahh its because the functions return the original named tuples so adding new ones doesn't work as a solution. Not sure how to fix this issue.

jorenham (commented):
Thanks! Besides the result type issue and the `python<3.12` CI failure, this looks very good to me.

scipy-stubs/stats/_binned_statistic.pyi
```diff
@@ -22,33 +23,59 @@ def binned_statistic(
...
     sample: onp.ToComplex2D,
     values: onp.ToComplex1D | Sequence[onp.ToComplex1D],
     statistic: _Statistic | Callable[[onp.ArrayND[np.float64]], onp.ToFloat] = "mean",
     bins: onp.ToInt | onp.ToFloat1D = 10,
     range: tuple[int, int] | None = None,
-    expand_binnumbers: bool = False,
     binned_statistic_result: BinnedStatisticddResult | None = None,
-) -> BinnedStatisticddResult: ...
+    *,
+    expand_binnumbers: Literal[True],
```
jorenham: This changes the order of the maybe-positional `binned_statistic_result` parameter. We can also also it keyword-only in this overload to avoid that.

PR #1532, opened 2026-04-11
✨ `fft`: `fftshift` and `ifftshift` shape-typing support

Aniketsy:
fixes #1106 

- add ndarray-preserving overloads for fftshift and ifftshift for input dtype and shape,

PR #1533, opened 2026-04-11
🐛 `fft`: `[i]fftshift` dtype preservation for int and bool inputs

Aniketsy:
discussed in #1532 comment.


jorenham:
Thanks @Aniketsy

PR #1534, opened 2026-04-11
🐛 `optimize`: generalize `newton` overloads to ND arrays

### LINKED ISSUE #1518: `optimize.newton` rejects functions that return an array
### Discussed in https://github.com/scipy/scipy-stubs/discussions/1517

<div type='discussions-op-text'>

<sup>Originally posted by **wilgri** April  7, 2026</sup>
Hello,

I'm using scipy-stubs and I'm new to optype. The following call of newton is flagged by basedpyright in strict mode:

```python
def func(x: ArrayND[np.float64]) -> ArrayND[np.float64]:
    return -2 * x + 1

newton(func, x0=np.zeros((2, 4), dtype=float))
```

I gives 

```
Type "(x: ArrayND[float64, AnyShape]) -> ArrayND[float64, AnyShape]" is not assignable to type "_Fun1D[ToFloat]"
   Function return type "ArrayND[float64, AnyShape]" is incompatible with type "ToFloat"
```

Looking at [newton's doc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.newton.html#newton) and in practice, `func` accepts ndarray of floats. What am I doing wrong ?

Looking at `scipy.optimize.newton` typing, I see that `func` can be a generic callable `_Fun1D` with a type variable named `_ToFloatT` . I'm not sure to understand it entirely but it seems that basedpyright does not find an appropriate overload and defaults `_ToFloatT` to `ToFloat` which is incompatible with `ArrayND[np.float64]`. Is my understanding correct ? In addition I don't understand why arguments are `onp.Array1D[np.float64]` instead of `ArrayND` ? I feel like I'm missing something.

How would you type this basic example ? 

</div>

fbourgey:
- Fixes https://github.com/scipy/scipy-stubs/issues/1518

- Update `optimize.newton` to preserve ND array shapes in its array overloads instead of just 1D inputs.

<details>

<summary>Python snippet with the different cases for `optimize.newton`</summary>

```python
import numpy as np
from numpy.typing import NDArray

from scipy.optimize import RootResults, newton


def scalar_func(x: float) -> float:
    return x * x - 2.0


def scalar_func_with_derivative(x: float) -> float:
    return x * x - 2.0


def scalar_derivative(x: float) -> float:
    return 2.0 * x


x0 = 1.0
scalar_root = newton(scalar_func, x0=x0)
scalar_root_full = newton(scalar_func, x0=x0, full_output=True)

print(scalar_root)
print(scalar_root_full)

def vector_func_1d(x: NDArray[np.float64]) -> NDArray[np.float64]:
    return -2.0 * x + 1.0


def vector_derivative_1d(x: NDArray[np.float64]) -> NDArray[np.float64]:
    return np.full_like(x, -2.0)


x0 = np.zeros(3, dtype=float)
vector_root_1d = newton(vector_func_1d, x0=x0)
vector_root_1d_full = newton(vector_func_1d, x0=x0, full_output=True)

print(vector_root_1d)
print(vector_root_1d_full)


def vector_func_2d(x: NDArray[np.float64]) -> NDArray[np.float64]:
    return -2.0 * x + 1.0


def vector_derivative_2d(x: NDArray[np.float64]) -> NDArray[np.float64]:
    return np.full_like(x, -2.0)


x0 = np.zeros((2, 4), dtype=float)
vector_root_2d = newton(vector_func_2d, x0=x0)
vector_root_2d_full = newton(vector_func_2d, x0=x0, full_output=True)

print(vector_root_2d)
print(vector_root_2d_full)
```

</details>

### DIFF
--- a/scipy-stubs/optimize/_zeros_py.pyi
+++ b/scipy-stubs/optimize/_zeros_py.pyi
@@ -20,10 +20,9 @@ _T = TypeVar("_T")
 _KT = TypeVar("_KT", bound=_FlagKey)
 _RT = TypeVar("_RT", bound=_Floating)
 _RT_co = TypeVar("_RT_co", bound=_Floating, default=_Float, covariant=True)
-_ToFloatT = TypeVar("_ToFloatT", bound=onp.ToFloat | onp.ToFloatND, default=onp.ToFloat)
-
+_ShapeT = TypeVar("_ShapeT", bound=tuple[int, ...])
 _Fun0D: TypeAlias = Callable[Concatenate[float, ...], onp.ToFloat] | Callable[Concatenate[np.float64, ...], onp.ToFloat]
-_Fun1D: TypeAlias = Callable[Concatenate[onp.Array1D[np.float64], ...], _ToFloatT]
+_FunND: TypeAlias = Callable[Concatenate[onp.Array[_ShapeT, np.float64], ...], onp.Array[_ShapeT, np.float64]]
 
 _State: TypeAlias = tuple[_FlagKey, _Float]
 _Bracket: TypeAlias = tuple[_Float, _Float]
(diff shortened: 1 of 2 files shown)


jorenham:
Nice, thanks @fbourgey

PR #1551, opened 2026-04-14
✅ signal: add tests for gauss_spline, spline_filter, qspline1d, qspline1d_eval, qspline2d, symiirorder1, symiirorder2

fbourgey:
Towards https://github.com/scipy/scipy-stubs/issues/1099

Covers: 

- `gauss_spline `
- `spline_filter`
- `qspline1d`
- `qspline1d_eval`
- `qspline2d`
- `symiirorder1`
- `symiirorder2`

### DIFF
--- a/tests/signal/test_spline_filters.pyi
+++ b/tests/signal/test_spline_filters.pyi
@@ -1,22 +1,29 @@
 # type-tests for `signal/_spline_filters.pyi`
 
-from typing import TypeAlias, assert_type
+from typing import Any, TypeAlias, assert_type
 
 import numpy as np
 import optype.numpy as onp
 import optype.numpy.compat as npc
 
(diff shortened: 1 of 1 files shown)


tests/signal/test_spline_filters.pyi
```diff
@@ -1,34 +1,70 @@
...
+    cspline2d,
+    gauss_spline,
+    qspline1d,
+    qspline1d_eval,
+    qspline2d,
+    spline_filter,
+    symiirorder1,
+    symiirorder2,
+)
+from scipy.signal._spline_filters import _ComplexQ, _FloatQ
```
jorenham: we should probably avoid importing these here, so we can be explicit about the expected return types that we're testing

fbourgey: 👍 I've removed them

jorenham:
Thanks!

Issue #406, opened 2025-01-03
`stats._multivariate`: generic shape-types for certain `*_frozen` types


jorenham:
in order of populatity (according to github code search):

- [x] `multivariate_normal_frozen`
- [ ] `multivariate_t_frozen`
- [x] `multinomial_frozen`
- [ ] `multivariate_hypergeom_frozen`
- [ ] `normal_inverse_gamma_frozen`
- [ ] `vonmises_fisher_frozen`
- [ ] `uniform_direction_frozen`
- [ ] `dirichlet_multinomial_frozen`
- [ ] `wishart_frozen`
- [ ] `invwishart_frozen`

Aniketsy:
hii @jorenham 😁 quick question on `matrix_normal_frozen ` as per docs state that `rowcov `and `colcov` accept a 1D array interpreted as diagonal entries, in addition to scalars and 2-D arrays. 
But as per my understanding we're missing 1D here, these are the diff , please correct me if I'm missing, or we should add. Thanks!
```
-        mean: onp.ToFloat2D | None = None,
-        rowcov: onp.ToFloat2D | onp.ToFloat = 1,
-        colcov: onp.ToFloat2D | onp.ToFloat = 1,
+        mean: onp.Array[tuple[_M, _N], npc.floating] | None = None,
+        rowcov: onp.ToFloat | onp.Array[tuple[_M, _M], npc.floating] = 1,
+        colcov: onp.ToFloat | onp.Array[tuple[_N, _N], npc.floating] = 1,
```

jorenham:
Hmm, if at runtime it also accepts 1d, then it should indeed also do so in the stubs. Did you check that it did? Because sometimes the docs are not entirely accurate in rare cases.

Aniketsy:
yes, it accepts 1d at runtime.

jorenham:
Ok, than you've uncovered a bug; nice :)

PR #1566, opened 2026-04-18
✨ `stats`: `multinomial` shape-typing

jorenham:
towards #406

ref: https://github.com/scipy/scipy-stubs/pull/1549#pullrequestreview-4134545369


jorenham:
cc @Aniketsy; this turned out to be way more complex that I had hoped... And since I don't expect it to be any simpler for the other multivariate distributions, it's absolutely fine (and maybe even wise) if you don't feel like working on these multivariate distributions.

Aniketsy:
> cc @Aniketsy; this turned out to be way more complex that I had hoped... And since I don't expect it to be any simpler for the other multivariate distributions, it's absolutely fine (and maybe even wise) if you don't feel like working on these multivariate distributions.

thanks, yes i can see quite large changes, so i'll not touch these as diving into these will create extra work for the review, but i'll try to understand how it is done to have some knowledge, and maybe in near future i can touch these ...
for now i'll try to pick simple and moderate level issues 😁(is this good way to move towards perfectionist ?  If not please correct me) , also please feel free to ping me if you feel any issue fits for me.

PR #1567, opened 2026-04-19
🔨 ignore `scipy.stats.chi2_contingency` in test coverage script

Aniketsy:
Towards #1099 

`scipy.stats.chi2_contingency` is a re-export of `scipy.stats.contingency.chi2_contingency`, which is already
tested in `tests/stats/test_contingency.pyi`.

### DIFF
--- a/scripts/test_coverage.py
+++ b/scripts/test_coverage.py
@@ -10,6 +10,12 @@ from typing import Final
 
 _SCIPY: Final = "scipy"
 _IGNORED_SUFFIXES: Final = {"__class__"}
+_IGNORED_QUALNAMES: Final = {
+    # `scipy.stats.chi2_contingency` is a re-export of
+    # `scipy.stats.contingency.chi2_contingency`, which is already tested in
+    # `tests/stats/test_contingency.pyi`.
+    "scipy.stats.chi2_contingency"
+}
 
 _PACKAGES_PUBLIC: Final = (
     "cluster",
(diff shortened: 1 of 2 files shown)


jorenham (commented):
`scipy.stats.chi2_contigency` is just a re-export of `scipy.stats.contingency.chi2_contingency`, for which we already have tests at https://github.com/scipy/scipy-stubs/blob/a912f40799fd8087c2f398baecbdc2524a375657/tests/stats/test_contingency.pyi. So I propose we either whitelist `scipy.stats.chi2_contigency` in the test coverage script, or move these tests into the existing test module for `stats.contingency`.

Aniketsy:
> So I propose we either whitelist `scipy.stats.chi2_contigency` in the test coverage script, or move these tests into the existing test module for `stats.contingency`.

thanks! for the review, which one should we prefer, i don't have much idea about whitelist `scipy.stats.chi2_contigency` in the test coverage script, but i'll be happy to explore and apply, as i'll learn something new.
 or if you will suggest with moving into existing i can do that . please let me know which will be best for codebase.

jorenham:
let's avoid code duplication then and ignore it in the script ([here](https://github.com/scipy/scipy-stubs/blob/master/scripts/test_coverage.py)), with an explanatory comment

scripts/test_coverage.py
```diff
@@ -225,6 +231,8 @@ def _parse_scipy_imports(tree: ast.AST) -> dict[str, str]:
 
 def _should_ignore(qualname: str) -> bool:
     """Check if a qualified name should be ignored (private or bare package)."""
+    if qualname in _IGNORED_QUALNAMES:
+        return True  # explicitly ignored public re-exports
```
jorenham: nitpick: this comment seems bit redundant

```suggestion
        return True
```

jorenham:
could you update the PR title and description?

jorenham:
thanks!

PR #1569, opened 2026-04-19
🐛 `stats`: Add missing 1D diagonal support for `matrix_normal` rowcov/colcov

Aniketsy:
follow up from #406 

> hii [@jorenham](https://github.com/jorenham) 😁 quick question on `matrix_normal_frozen ` as per docs state that `rowcov `and `colcov` accept a 1D array interpreted as diagonal entries, in addition to scalars and 2-D arrays. But as per my understanding we're missing 1D here, these are the diff , please correct me if I'm missing, or we should add. Thanks!
> 
> ```
> -        mean: onp.ToFloat2D | None = None,
> -        rowcov: onp.ToFloat2D | onp.ToFloat = 1,
> -        colcov: onp.ToFloat2D | onp.ToFloat = 1,
> +        mean: onp.Array[tuple[_M, _N], npc.floating] | None = None,
> +        rowcov: onp.ToFloat | onp.Array[tuple[_M, _M], npc.floating] = 1,
> +        colcov: onp.ToFloat | onp.Array[tuple[_N, _N], npc.floating] = 1,
> ```

### DIFF
--- a/scipy-stubs/stats/_multivariate.pyi
+++ b/scipy-stubs/stats/_multivariate.pyi
@@ -234,8 +234,8 @@ class matrix_normal_gen(multi_rv_generic):
         self,
         /,
         mean: onp.ToFloat2D | None = None,
-        rowcov: onp.ToFloat2D | onp.ToFloat = 1,
-        colcov: onp.ToFloat2D | onp.ToFloat = 1,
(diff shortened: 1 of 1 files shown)


jorenham:
Thanks !

PR #1574, opened 2026-04-21
👷ci: skip type-check and stubtest jobs when unrelated files change

### LINKED ISSUE #608: CI: only type-check and stubtest if needed
Stubtest is only needed when there are changes to `scipy-stubs/**/*.pyi`, `.mypyignore`, `uv.lock` or `pyproject.toml`. The type-checkers should also be run if there are changes to a `*.py` or `*.pyi` in general.

Aniketsy:
fixes #608

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: detect changes
         id: filter
-        uses: dorny/paths-filter@v4.0.1
+        uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d  # v4.0.1
         with:
           filters: |
             typecheck:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,30 @@ env:
 # UV_LOCKED: 1  # TODO(@jorenham): re-enable after restoring `UV_PRERELEASE`
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      typecheck: ${{ steps.filter.outputs.typecheck }}
+      stubtest: ${{ steps.filter.outputs.stubtest }}
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - name: detect changes
+        id: filter
+        uses: dorny/paths-filter@v4.0.1
+        with:
+          filters: |
+            typecheck:
+              - '**/*.py'
+              - '**/*.pyi'
+            stubtest:
+              - 'scipy-stubs/**/*.pyi'
+              - '.mypyignore'
+              - 'uv.lock'
+              - 'pyproject.toml'
+
   lint:
     runs-on: ubuntu-latest
     timeout-minutes: 2
@@ -73,6 +97,8 @@ jobs:
         run: uv run stubdefaulter --check --exit-zero --packages .
 
   typecheck:
+    needs: changes
+    if: ${{ needs.changes.outputs.typecheck == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
@@ -110,9 +136,12 @@ jobs:
         run: mypy --no-incremental --cache-dir=/dev/null scipy-stubs
 
       - name: stubtest scipy
+        if: ${{ needs.changes.outputs.stubtest == 'true' }}
         run: stubtest --ignore-disjoint-bases --allowlist=.mypyignore scipy
 
   typetest:
+    needs: changes
+    if: ${{ needs.changes.outputs.typecheck == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       typecheck: ${{ steps.filter.outputs.typecheck }}
       stubtest: ${{ steps.filter.outputs.stubtest }}
     steps:
-      - uses: actions/checkout@v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: detect changes
         id: filter
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,14 @@ permissions: read-all
 
 on:
   pull_request:
+    paths:
+      - '**/*.py'
+      - '**/*.pyi'
   push:
     branches: [master]
+    paths:
+      - '**/*.py'
+      - '**/*.pyi'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -13,36 +12,55 @@ concurrency:
...
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    outputs:
+      typecheck: ${{ steps.filter.outputs.typecheck }}
+      stubtest: ${{ steps.filter.outputs.stubtest }}
+    steps:
+      - uses: actions/checkout@v6.0.2
```
Aniketsy:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```

.github/workflows/ci.yml
```diff
@@ -13,36 +12,55 @@ concurrency:
...
+
+      - name: detect changes
+        id: filter
+        uses: dorny/paths-filter@v4.0.1
```
Aniketsy:
```suggestion
        uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d  # v4.0.1
```

Aniketsy:
I've looked into these erorrs, we were getting and got zizmor requires SHA instead of tag, for the security purpose, this is something new for me . @jorenham could you please explain me this, if the things i got from the search are correct, i'd like to learn more about this . Thankyou 😊

jorenham:
> I've looked into these erorrs, we were getting and got zizmor requires SHA instead of tag, for the security purpose, this is something new for me . @jorenham could you please explain me this, if the things i got from the search are correct, i'd like to learn more about this . Thankyou 😊

It's a guard against supply chain attacks: https://github.blog/security/supply-chain-security/safeguard-container-signing-capability-actions/

.github/workflows/ci.yml
```diff
@@ -16,6 +16,30 @@ env:
...
+          filters: |
+            typecheck:
+              - '**/*.py'
+              - '**/*.pyi'
+            stubtest:
+              - 'scipy-stubs/**/*.pyi'
+              - '.mypyignore'
+              - 'uv.lock'
+              - 'pyproject.toml'
+
```
jorenham: What's the advantage of this action, as opposed to just using the `paths` like we do at numpy (https://github.com/numpy/numpy/blob/9546c7fe4d81a0a02ae224aa60a3c493db2ba92a/.github/workflows/stubtest.yml#L13-L18) ?

Aniketsy: main reason of using this as we have two conditions,` typecheck` and `stubtest` inside the same workflow, and as per what i got for using `path`  we can include or exclude one condition for the whole workflow  https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore  please let me know your thoughts on this , if i misunderstood anything.

jorenham: I prefer splitting it up then. Every external action increases the risk of supply chain attacks (I've gotten a bit paranoid after the recent supply chain attacks).

Aniketsy: sure, yes, splitting will be better and safe .

Aniketsy:
In the new file, i mostly copied from the `ci.yml` , i dont think we should get these errors . could you please help with this and apologize for creating extra work please let me know if splitting should be done as i approached. Thanks!

jorenham:
Pinning the actions to their full-length commit hash should resolve the zizmor errors. Just look at any other workflow for examples of this.

Aniketsy:
> Pinning the actions to their full-length commit hash should resolve the zizmor errors. Just look at any other workflow for examples of this.

yes , i got this `- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2` ,  i think this will fix, as previously i searched for this, but in our workflow we haven't used this , so should we update in other files too or its fine

jorenham:
> but in our workflow we haven't used this , so should we update in other files too or its fine

what workflow do you mean?

Aniketsy:
> > but in our workflow we haven't used this , so should we update in other files too or its fine
> 
> what workflow do you mean?

i was looking into other files, like `update_coverage.yml`, `mypy_primer.yml`

jorenham:
> > > but in our workflow we haven't used this , so should we update in other files too or its fine
> > 
> > 
> > what workflow do you mean?
> 
> i was looking into other files, like `update_coverage.yml`, `mypy_primer.yml`

Those also have pinned actions though?
- https://github.com/scipy/scipy-stubs/blob/master/.github/workflows/mypy_primer.yml
- https://github.com/scipy/scipy-stubs/blob/master/.github/workflows/update_coverage.yml

Aniketsy:
> > > > but in our workflow we haven't used this , so should we update in other files too or its fine
> > > 
> > > 
> > > what workflow do you mean?
> > 
> > 
> > i was looking into other files, like `update_coverage.yml`, `mypy_primer.yml`
> 
> Those also have pinned actions though?
> 
> * https://github.com/scipy/scipy-stubs/blob/master/.github/workflows/mypy_primer.yml
> * https://github.com/scipy/scipy-stubs/blob/master/.github/workflows/update_coverage.yml

ahh i guess my fault, i'll be behind some commits, thats the reason i can see older version. Thanks!

.github/workflows/stubtest.yml
```diff
@@ -37,10 +37,10 @@ jobs:
         np: ["1.26", "2.0", "2.1", "2.2", "2.3", "2.4"]
 
     steps:
-      - uses: actions/checkout@v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
jorenham, line 40:
```suggestion
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          persist-credentials: false
```

jorenham:
Hmm, sorry, but I don't think we should do this after all. Because as you can see, this would mean that we'll no longer be able to require the typecheck/typetest workflows to pass, which would break the github auto-merge feature.
It's good that we know this now though, so I hope it doesn't feel like you've wasted your time on this. Thanks for your work.

Aniketsy:
>  Because as you can see, this would mean that we'll no longer be able to require the typecheck/typetest workflows to pass, which would break the github auto-merge feature. It's good that we know this now though, so I hope it doesn't feel like you've wasted your time on this. Thanks for your work.

oh, we got to know new thing, and no issues 😊 , i don't feel like wasted my time, as i got to learn many thing, how to split workflow, about how to safe guard from supply chain attacks, these are new things for me.

Thanks for the review in every step and guiding me along the way ❤️ .

PR #1583, opened 2026-05-03
🐛 `sparse.csgraph`: add missing callable return type for `laplacian` with `form="function"`

### LINKED ISSUE #1580: `sparse.csgraph.laplacian`: missing callable return type for `form="function"`
`sparse.csgraph.laplacian` (https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.laplacian.html)
```
python3 - << 'EOF'
import numpy as np
from scipy.sparse.csgraph import laplacian
from scipy.sparse.linalg import LinearOperator
from scipy.sparse._base import _spbase

G = np.array([[0,1,1],[1,0,1],[1,1,0]], dtype=float)
fn = laplacian(G, form="function")

print("type:", type(fn))
print("callable:", callable(fn))
print("is _spbase:", isinstance(fn, _spbase))
print("is LinearOperator:", isinstance(fn, LinearOperator))
print("is ndarray:", isinstance(fn, np.ndarray))
EOF
```

output:
```
type: <class 'function'>
callable: True
is _spbase: False
is LinearOperator: False
is ndarray: False

```

also we can have small improvements in the docs, as the top line does not mention function, but it is in description . (in my opinion we should add in top line also).
i feel we have some more gaps in our stubs, but i need to verify, 

@jorenham please let me know, your thoughts on this .

Aniketsy:
fixes #1580

### DIFF
--- a/tests/sparse/csgraph/test_laplacian.pyi
+++ b/tests/sparse/csgraph/test_laplacian.pyi
@@ -1,0 +1,19 @@
+from typing import reveal_type
+
+import numpy as np
+import numpy.typing as npt
+
+from scipy.sparse.csgraph import laplacian
+
+G: npt.NDArray[np.float64]
+v: npt.NDArray[np.float64]
+
+fn = laplacian(G, form="function")
+reveal_type(fn)  # Callable
+
+result = fn(v)
+reveal_type(result)
+
+fn2, diag = laplacian(G, form="function", return_diag=True)
+reveal_type(fn2)
+reveal_type(diag)
--- a/tests/sparse/csgraph/test_laplacian.pyi
+++ b/tests/sparse/csgraph/test_laplacian.pyi
@@ -1,19 +1,18 @@
-from typing import reveal_type
+from collections.abc import Callable
 
 import numpy as np
 import numpy.typing as npt
+import optype as op
+import optype.numpy as onp
+import optype.numpy.compat as npc
 
 from scipy.sparse.csgraph import laplacian
 
-G: npt.NDArray[np.float64]
-v: npt.NDArray[np.float64]
+_f64_nd: npt.NDArray[np.float64]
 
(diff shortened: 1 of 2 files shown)


jorenham (changes requested):
If we remove the `"function" from  `_Form`, we no longer have overlapping overloads, allowing us to *only* return the callable for `form="function"`, i.e. so that in case of "array" and "lo" the return type will stay the same as it currently is.

tests/sparse/csgraph/test_laplacian.pyi
```diff
@@ -0,0 +1,19 @@
+from typing import reveal_type
+
+import numpy as np
+import numpy.typing as npt
+
+from scipy.sparse.csgraph import laplacian
+
+G: npt.NDArray[np.float64]
+v: npt.NDArray[np.float64]
```
jorenham: I prefer the usual naming convention for these, i.e. using something like `_f64_nd` for both

tests/sparse/csgraph/test_laplacian.pyi
```diff
@@ -0,0 +1,19 @@
...
+
+fn = laplacian(G, form="function")
+reveal_type(fn)  # Callable
+
+result = fn(v)
+reveal_type(result)
+
+fn2, diag = laplacian(G, form="function", return_diag=True)
+reveal_type(fn2)
+reveal_type(diag)
```
jorenham: `reveal_type` doesn't test anything; it's purely a debugging tool; use `assert_type` or `optype.test.assert_subtype` ([docs](https://jorenham.github.io/optype/reference/experimental/test/)).

Aniketsy: ahh, yes  i'll add `optype.test.assert_subtype` for this.

jorenham (approved):

PR #1592, opened 2026-05-12
🐛: special: Refine overloads for multiufuncs

fbourgey:
Follow-up of https://github.com/scipy/scipy-stubs/pull/159 following https://github.com/scipy/scipy-stubs/pull/1590#issuecomment-4416103534


jorenham:
supplanted by #1606

Issue #1614, opened 2026-05-26
Tracker: SciPy 1.18.0 support


jorenham:
https://github.com/scipy/scipy/releases/tag/v1.18.0rc1

- [x] Bump SciPy and Python - #1615
  - Add temp ignores (ruff, stubtest, unstubbed modules)
  - CI matrix updates (+ update required gh checks)
  - TOX
  - Version literals in stubs
  - Docs (i.e. `*.md` references to now-dropped versions)
- [x] Fix stubtest errors (resolve `.mypyignore` todo's)
- [x] Fix `scripts/unstubbed_modules.py` errors (resolve `TODO_1_18`)
- [x] Fix ruff errors (pyupgrade; see `# TODO` in `pyproject.toml` under `[tool.ruff.lint]`)
  - [x] `UP040`
  - [x] `UP046`
  - [x] `UP047`
- [x] New deprecations (https://github.com/scipy/scipy/releases/tag/v1.18.0rc1)
- [x] After stubtest: Manually check release notes for any other changes
  - [x] `sparse.linalg.LinearOperator` N-d support (use generic shape type)
  - [x] `sparse.linalg.minres` complex support
  - [x] `spatial.distance.{minkowski, euclidean, seuclidean}` N-d support
  - [x] `spatial.KDTree.sparse_distance_matrix` *can* now return sparse *arrays*
- [x] replace `np.bool_` with `np.bool`

Aniketsy:
hi @jorenham  please let me know if you're working on these or i can help with this..
- Fix stubtest errors (resolve .mypyignore todo's)
- Fix ruff errors (pyupgrade; see # TODO in pyproject.toml under [tool.ruff.lint])

jorenham:
> hi [@jorenham](https://github.com/jorenham) please let me know if you're working on these or i can help with this..
> 
> * Fix stubtest errors (resolve .mypyignore todo's)
> * Fix ruff errors (pyupgrade; see # TODO in pyproject.toml under [tool.ruff.lint])

There are some annoying things related to PEP 695 that need to be worked around. And because we have a deadline, it's better if I do both. But thanks for the offer :)

PR #1615, opened 2026-05-26
📌 scipy 1.18.0rc1

jorenham:
This is but the first step in the big 1.17 -> 1.18 metamorphosis, and ignores almost all of the new errors (for now, of course).

This also drops Python 3.11 support, because SciPy 1.18.0 also no longer supports it. This means that PEP 695 syntax is now an option (but because of implicit variance probably not always a good idea to use).

Towards #1614


jorenham:
GitHub Actions is having issues (again): https://www.githubstatus.com/:

> **Investigating** - We are investigating reports of degraded performance for Actions and Pages
> May 26, 2026 - 10:57 UTC

PR #1624, opened 2026-05-26
👽️ `linalg.bandwidth`: sync with upstream (backwards incompatible type change)

jorenham:
Several upstream changes for `scipy.linalg.bandwith`:

- The implementation moved from `_cythonized_array_utils` to `linalg._misc` 
  - No problems here
- Batching for >2-d input
  - Not mentioned in the [rc1 relnotes](https://github.com/scipy/scipy/releases/tag/v1.18.0rc1)
- For 2-D input, the return type changed from `(int, int)` to `(np.int64, np.int64)`
  - Also not mentioned in the [rc1 relnotes](https://github.com/scipy/scipy/releases/tag/v1.18.0rc1).
  - From a typing perspective, this is a **backwards-incompatible breaking change**, because even though `np.int64` and `int` are mostly ducktype-compatible, they are incompatible nominal types. This is likely to lead to new typing errors, because  `a: int = np.int64(1)` is invalid, and so is `b: np.int64 = 1`. So "swapping" `int` for `int64` or vice-versa won't be accepted by type-checkers [^1]. 

I think it would be best to revert the `int`/`np.int64` return type change. The new batching support is probably also worth mentioning.

Towards #1614

[^1]: For `float`/`np.float64` this wouldn't have been an issue, because unlike `int`/`np.int64`, `np.float64` subclasses `builtins.float`. So `x: float = np.float64(0.78)` is accepted by all type-checkers.

---

**Update**: I did a quick github code search ([repro](https://github.com/search?q=%2Flinalg%5C.bandwidth%5C%28%2F+language%3APython++NOT+is%3Aarchived+NOT+is%3Afork&type=code)), and literally the first hit is an example of where this will cause typing errors: https://github.com/NatLabRockies/batmods-lite/blob/7fdb441d678df459a87834b32e6dedc3dc4cee12/src/bmlite/_core/_idasolver.py#L111

### DIFF
--- a/.mypyignore
+++ b/.mypyignore
@@ -20,8 +20,6 @@ scipy\.stats\.(_new_distributions\.)?Normal\.__new__
 ###
 # TODO: 1.18.0rc1
 
-scipy.linalg._cythonized_array_utils.__all__
-scipy.linalg._cythonized_array_utils.bandwidth
 scipy.linalg._decomp.eigvals
 scipy.linalg._decomp_lu_cython
 scipy.linalg._matfuncs.logm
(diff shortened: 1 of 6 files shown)


jorenham:
cc @tylerjereddy

tylerjereddy:
Noted in my lab notes this evening, thanks Joren

PR #1632, opened 2026-05-27
👽️ `spatial.transform`: upstream changes

jorenham:
The `RigidTransform` and `Rotation` classes are now subscriptable at runtime (i.e. the runtime equivalent of generic types). The relnotes don't mention this, but it might be nice if they would.

They can now also be composed with one another using the `*` infix operator. This is mentioned in the relnotes.

towards #1614


scipy-stubs/spatial/transform/_rigid_transform.pyi
```diff
@@ -82,17 +89,31 @@ class RigidTransform(Generic[_ShapeT_co]):
...
-    def __mul__(self: RigidTransform[_JustAnyShape], other: RigidTransform, /) -> RigidTransform: ...
+    def __rmul__(self: RigidTransform[_JustAnyShape], other: Rotation[tuple[int, ...]], /) -> RigidTransform: ...
     @overload
-    def __mul__(self: RigidTransform[tuple[()]], other: RigidTransform[_ShapeT], /) -> RigidTransform[_ShapeT]: ...
+    def __rmul__(self: RigidTransform[tuple[()]], other: Rotation[_ShapeT], /) -> RigidTransform[_ShapeT]: ...
     @overload
-    def __mul__(self: RigidTransform[tuple[int]], other: RigidTransform[tuple[int]], /) -> RigidTransform[tuple[int]]: ...
+    def __rmul__(self: RigidTransform[tuple[int]], other: Rotation[tuple[int]], /) -> RigidTransform[tuple[int]]: ...
     @overload
-    def __mul__(self, other: RigidTransform, /) -> RigidTransform: ...  # pyright: ignore[reportOverlappingOverload]
```
jorenham, line 95: The fact that this pyright error is no longer reported demonstrates that it was indeed a false positive (or a false negative right now, but I highly doubt that, since all 3 other type-checkers don't complain here).

PR #1641, opened 2026-05-27
✅ sparse : add tests for `scipy.sparse.csgraph`

Aniketsy:
towards #1099

### DIFF
--- a/tests/sparse/csgraph/test_flow.pyi
+++ b/tests/sparse/csgraph/test_flow.pyi
@@ -1,0 +1,13 @@
+from typing import assert_type
+
+import numpy as np
+
+import scipy.sparse as sparse
+from scipy.sparse.csgraph import maximum_flow
+
+csr_arr: sparse.csr_array[np.float64, tuple[int, int]]
+
+flow = maximum_flow(csr_arr, 0, 1)
+
+assert_type(flow.flow_value, int | np.int32 | np.int64)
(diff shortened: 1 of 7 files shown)


jorenham:
Sould you rebase onto main? Some `csgraph` code changed.

tests/sparse/csgraph/test_flow.pyi
```diff
@@ -0,0 +1,13 @@
+from typing import assert_type
+
+import numpy as np
+
+import scipy.sparse as sparse
+from scipy.sparse.csgraph import maximum_flow
+
+csr_arr: sparse.csr_array[np.float64, tuple[int, int]]
+
+flow = maximum_flow(csr_arr, 0, 1)
```
jorenham: it helps avoid IDEs from suggesting these as imports (which clearly isn't what we wanbt) by prefixing them with `_` (same for the other test modules in this PR

jorenham (changes requested):
could you use the same `###` comment style to separate the test-data and the individual tested functions?

Aniketsy:
from next onwards i'll make sure to not include, many tests in one PR, so that i think it will be easy to review. in this i have changed the format to all `csgraph test`

>could you use the same ### comment style to separate the test-data and the individual tested functions?

jorenham (approved):

jorenham:
Thanks!

PR #1593, opened 2026-05-12
⬆️ pyrefly 1.0.0

jorenham:
https://github.com/facebook/pyrefly/releases/tag/1.0.0

supplants #1589

PR #1649, opened 2026-05-28
MNT: update pyrefly to 1.0.0

loicdiridollou:


jorenham:
Thanks!

... but this seems to be a duplicate of #1593. And as I said [there](https://github.com/scipy/scipy-stubs/pull/1593#issuecomment-4517395604); we'll have to wait until the next pyrefly release (because of the `bad-specialization` error).

So I'm afraid I'm going to have to close this.

PR #1680, opened 2026-05-31
👽️ `linalg.cho_factor`: change 2nd return type to bool array

jorenham:
The second return type of `scipy.linalg.cho_factor` silently changed from `bool` to `NDArray[np.bool]`, even if 0d (which may or may not have been intentional).
Either way, it should probably be mentioned in the release notes, because this could be a breaking change in some cases.


jorenham:
@tylerjereddy this sounds like something you'd want to know about

tylerjereddy:
thanks, added to my list for RC2

PR #1681, opened 2026-05-31
👽️ `interpolate.splint`: change 2nd return type to `None` (`full_output=True`)

jorenham:
The type of `scipy.interpolate.splint(..., full_output=True)[1]` changed from a 1-d `float64` array to `None`.

I'm guessing this is because of the fortran -> C transition, but it's a breaking change nonetheless, and might be worth mentioning the relnotes.

I guess there's also something to be said for deprecating `full_output` parameter.


jorenham:
@tylerjereddy

tylerjereddy:
Added to my list for RC2, and maybe cc @ev-br just in case it is unexpected

ev-br:
Ouch, thanks for the ping. Yes, technically a backwards compat break, most likely a forgotten shortcut during porting from Fortran to C. That said, the missing functionality is a _very _  niche multipack era remnant, and most likely has exactly zero users.

PR #1683, opened 2026-05-31
👽️ `stats.binomtest`: batching support (undocumented)

jorenham:
The types of the `k` and `n` attributes of the reutrned  `BinomTestResult` instance changed from `int` to `np.float64`, even if you pass `k` and `n` as `int`. Technically speaking, this is a breaking change, but I'm guessing that this was an oversight.

`scipy.stats.binomtest` now also supports batching of `k` and `n` and `p`. This is undocumented (not even in the docstring), and not (directly) mentioned in the release notes.

### DIFF
--- a/scipy-stubs/stats/_binomtest.pyi
+++ b/scipy-stubs/stats/_binomtest.pyi
@@ -1,21 +1,55 @@
 from types import ModuleType
-from typing import Literal
+from typing import Generic, Literal, overload
+from typing_extensions import TypeVar
+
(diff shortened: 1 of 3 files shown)


jorenham:
@tylerjereddy found another one :sweat_smile:

tylerjereddy:
added to my list for RC2, thanks, and

> but I'm guessing that this was an oversight.

cc @mdhaber just in case he wants me to do something about it re: rel process

Issue #1673, opened 2026-05-30
`stats.sampling.RatioUniforms`: shape-typing


jorenham:
`rvs` currently returns ~`np.float64 | onp.ArrayND[np.float64]`~ `onp.ArrayND[np.float64]`, which could be overloaded to improve this

PR #1685, opened 2026-06-02
✨ `stats`: Narrow `RatioUniforms.rvs` return type to array of `np.float64`

loicdiridollou:
Towards #1673

Documentation did not mention the case for returning `np.float64`, only an array of `np.float64` (https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.sampling.RatioUniforms.rvs.html#scipy.stats.sampling.RatioUniforms.rvs)


scripts/scipy_usage.py
```diff
@@ -373,7 +373,7 @@ def visit_Call(self, node: ast.Call) -> None:
             (call_name := self._get_call_name(node.func))
             and (scipy_call := self._resolve_scipy_call(call_name))
         ):  # fmt: skip
-            self.calls.add(scipy_call)  # ty:ignore[possibly-unresolved-reference]  # false positive
```
jorenham, line 376: seems unrelated

tests/stats/test_resampling.pyi
```diff
@@ -86,3 +87,10 @@ assert_type(PermutationMethod(n_resamples=999), PermutationMethod)
 
 # power
 assert_type(power(_statistic_1d, np.random.standard_normal, [10, 20]), PowerResult[Any])
+
+# RationUniforms.rvs
+v = np.sqrt(np.exp(np.sqrt(2))) * np.sqrt(2)
+umax = np.sqrt(np.exp(0))
+gen = RatioUniforms(np.exp, umax=umax, vmin=-v, vmax=v)
+assert_type(gen.rvs(), onp.ArrayND[np.float64])
+assert_type(gen.rvs(size=1234), onp.ArrayND[np.float64])
```
jorenham: `RatioUniforms` is already tested in `test_sampling`, so this can be reverted

jorenham (changes requested):
The SciPy docs aren't always accurate when it comes to types. But I checked it myself, and I indeed wasn't able to get `.rvs` to return a bare `float64` scalar. Perhaps it changed somewhere between releases or something. 

Either way, good catch; thanks!

BTW, the PR title doesn't reflect the changes. This isn't about shape-typing. This just narrows the return type to make it more accurate.

jorenham (approved):

jorenham:
Thanks @loicdiridollou

Issue #1445, opened 2026-03-15
`sparse.csgraph.laplacian`: improve return dtypes


jorenham:
it can currently return things like `onp.Array2D[npc.number] | _spbase | LinearOperator`, which is very awkward

PR #1688, opened 2026-06-03
✨ `sparse.csgraph`: Refine `laplacian` return type

loicdiridollou:
Towards #1445


scipy-stubs/sparse/csgraph/_laplacian.pyi
jorenham, line 1: Comments that briefly explain what the overload covers is something we do quite a bit in scipy-stubs. But we use a specific commenting style that would be nice if you could also adhere to here. See https://github.com/scipy/scipy-stubs/blob/a4504b2faf6f4668f138fa9df4614b7a86f106c5/scipy-stubs/linalg/_decomp_lu.pyi for example.

jorenham (changes requested):
Thanks! The type alias changes make it a bit difficult for me to see what exactly changed here and what the consequences of these changes are. For that it would help if you could rebase onto main, because in the meantime type-tests have been added for `sparse.csgraph.laplacian`, which could help tell us if this leads to any regressions.

scipy-stubs/sparse/csgraph/_laplacian.pyi
loicdiridollou: Thanks for pointing me to another example, tried to improve the comments, let me know if it makes more sense.

tests/sparse/csgraph/test_laplacian.pyi
```diff
@@ -21,6 +21,9 @@ _f64_nd: npt.NDArray[np.float64]
...
 
-_fn2, _diag = laplacian(_f64_nd, form="function", return_diag=True)
-op.test.assert_subtype[Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]](_fn2)
-op.test.assert_subtype[onp.Array1D[npc.number]](_diag)
+fn1 = laplacian(_f64_nd, form="lo")
+op.test.assert_subtype[Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]](fn1)
+
+fn2, diag = laplacian(_f64_nd, form="function", return_diag=True)
+op.test.assert_subtype[Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]](fn2)
+op.test.assert_subtype[onp.Array1D[npc.number]](diag)
```
jorenham: It helps to prefix these temp variables with `_` so that they don't show up in auto-import completions of certain IDEs, i.e. avoiding namespace pollution.

scipy-stubs/sparse/csgraph/_laplacian.pyi
jorenham, line 1: Could you place the per-overload comments directly after the `@overload` decorator? That way they're always aligned, which makes it a bit easier to read

jorenham (changes requested):

scipy-stubs/sparse/csgraph/_laplacian.pyi
loicdiridollou: Pushed an update, let me know if this is where you expected the comments to be

jorenham (changes requested):
The pending CI jobs are because of the new NumPy 2.5 release jobs I recently added, so if you could rebase then we can also test against that here.

The `form="lo"` case is now incorrect: it should return `LinearOperator` (with appropriate shape (so 2d I think) and dtype), but now it would return a plain `Callable`. 
So the type test you added now incorrectly asserts that "form=lo" should return a callable, but that should be a `LinearOperator`.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
     symmetrized: bool = False,
-) -> _LaplacianMatrix: ...
+) -> _LaplacianLO: ...
 @overload
-def laplacian(
+def laplacian(  # lo form + return_diag -> LinearOperator + diag
     csgraph: _ToCSGraph,
+    normed: bool = False,
+    *,
+    return_diag: onp.ToTrue,
```
jorenham, line 62: The `return_diag` param can also be used positioally, and since it's the third param, that's a pretty likely to be done in practice. But this changes it so that passing `True` for `return_diag` positionally will no longer be accepted; so that's a regression. The solution is to have another overload like this one that handles the positional `return_diag` case, i.e. where `normed` doesn't have a default.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
+    csgraph: _spbase,
+    normed: bool = False,
+    return_diag: onp.ToFalse = False,
+    use_out_degree: bool = False,
+    *,
+    copy: bool = True,
+    form: Literal["array"] = "array",
+    dtype: npt.DTypeLike | None = None,
+    symmetrized: bool = False,
+) -> _spbase: ...
```
jorenham, line 116: Can we narrow this (and in the two overloads below)? returning a bare `_spbase` doesn't really rhyme with the intent of this PR.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
 import optype.numpy.compat as npc
 
 from scipy.sparse._base import _spbase
-from scipy.sparse.linalg import LinearOperator
+from scipy.sparse.linalg._interface import LinearOperator
```
jorenham, line 9: ?

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
 
 ###
 
 type _LaplacianFunction = Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]
-type _LaplacianMatrix = onp.Array2D[npc.number] | _spbase | LinearOperator
 type _LaplacianDiag = onp.Array1D[npc.number]
+type _LaplacianLO = LinearOperator[npc.number, tuple[int, int]]
 type _ToCSGraph = onp.ToComplex2D | _spbase
-type _Form = Literal["array", "lo"]
+type _FunctionForm = Literal["function"]
```
jorenham, line 17: Why does only "function" have a type alias, and not also the other "lo" and "array" literals? I don't care if you use type aliases or inline the `Literal`s, as long as it's consistent.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
 from scipy.sparse._base import _spbase
-from scipy.sparse.linalg import LinearOperator
+from scipy.sparse.linalg._interface import LinearOperator
 
 ###
 
 type _LaplacianFunction = Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]
-type _LaplacianMatrix = onp.Array2D[npc.number] | _spbase | LinearOperator
 type _LaplacianDiag = onp.Array1D[npc.number]
+type _LaplacianLO = LinearOperator[npc.number, tuple[int, int]]
```
jorenham, line 15: The `LinearOperator` without generic arguments also accepts `np.bool_`, but now it doesn't anymore, so this narrows the dtype. Did you verify that at runtime `np.bool_`-values `LinearOperator`s are indeed rejected? Because if there aren't (i.e. accepted) then this should also accept `np.bool_`.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
+    return_diag: onp.ToTrue,
+    use_out_degree: bool = False,
+    copy: bool = True,
+    form: Literal["lo"],
+    dtype: npt.DTypeLike | None = None,
+    symmetrized: bool = False,
+) -> tuple[_LaplacianLO, _LaplacianDiag]: ...
+@overload
+def laplacian(  # array form dense input -> dense output
+    csgraph: onp.ToComplex2D,
```
jorenham, line 71: This broadens the accepted `csgraph` types in a type-unsafe way, because at runtime passing a nested sequence of `float` or something will crash.

tests/sparse/csgraph/test_laplacian.pyi
```diff
@@ -21,6 +22,9 @@ _f64_nd: npt.NDArray[np.float64]
 _fn = laplacian(_f64_nd, form="function")
 op.test.assert_subtype[Callable[[onp.ToComplex2D], onp.Array2D[npc.number]]](_fn)
 
+_fn1 = laplacian(_f64_nd, form="lo")
+op.test.assert_subtype[LinearOperator[npc.number, tuple[int, int]]](_fn1)
```
jorenham, line 26: This doesn't cover all overloads that you added.

scipy-stubs/sparse/csgraph/_laplacian.pyi
```diff
@@ -6,75 +6,135 @@ import optype.numpy as onp
...
-    form: _Form = "array",
+    form: Literal["array"] = "array",
     dtype: npt.DTypeLike | None = None,
     symmetrized: bool = False,
-) -> tuple[_LaplacianMatrix, _LaplacianDiag]: ...
+) -> tuple[onp.Array2D[npc.number], _LaplacianDiag]: ...
 @overload
-def laplacian(
-    csgraph: _ToCSGraph,
+def laplacian(  # array form dense input + return_diag -> dense output + diag
```
jorenham, line 94: the comment here is the same as the one above, which tbh kinda defeats the purpose

jorenham (changes requested):

jorenham:
this seems to have stranded, so I'm going to close this in favor of #2095

PR #1696, opened 2026-06-04
🐛✨ `stats.make_distribution`: shape-typing support

### LINKED ISSUE #1689: `CustomContinuousDistribution` parameters can be arrays
```py
class UnivariateDistribution(_BaseDistribution[_XT_co], Generic[_XT_co, _ShapeT0_co]):
    __array_priority__: ClassVar = 1
    _parameterizations: ClassVar[Sequence[_Parameterization]]
    _not_implemented: Final[str]
    _original_parameters: dict[str, _XT_co | onp.ArrayND[_XT_co, _ShapeT0_co]]
    _variable: _Parameter

    def __init__(
        self,
        *,
        tol: float | _Null | None = ...,
        validation_policy: _ValidationPolicy = None,
        cache_policy: _CachePolicy = None,
        **parameters: onp.ToFloat | None,
    ) -> None: ...
```
Here: `parameters` can be an array.

jorenham:
Pfew, that wasn't easy... 

I had to jump through some hoops to work around the lack of higher-kinded typing (HKT) support in Python using heavily overloaded type-check-only protocols to emulate generic type constructors. It seems to work for pyrefly and (based)pyright, but mypy seems to get confused somewhere in the overload spaghetti, and sometimes infers a 0-d array instead of a bare scalar type (which isn't a very big difference, so it's probably fine to just ignore it).

This also replaces a bunch of awkward `np.floating` return types with `np.float64`. Technically it's not correct, but in 99/100 cases people use `float64` in my experience, so it's probably fine. We can reconsider broadening again if this turns out to be an issue for someone in practice.

Closes #1689

### DIFF
--- a/scipy-stubs/stats/_distribution_infrastructure.pyi
+++ b/scipy-stubs/stats/_distribution_infrastructure.pyi
@@ -37,10 +37,10 @@ __all__ = ["Mixture", "abs", "exp", "log", "make_distribution", "order_statistic
 
 _Tss = ParamSpec("_Tss", default=...)
 
-_FloatT = TypeVar("_FloatT", bound=_Float, default=_Float)
-_FloatT_co = TypeVar("_FloatT_co", bound=_Float, default=_Float, covariant=True)
-_IntT_co = TypeVar("_IntT_co", bound=_Int, default=_Int, covariant=True)
-_RealT_co = TypeVar("_RealT_co", bound=_Float | _Int, default=_Float | _Int, covariant=True)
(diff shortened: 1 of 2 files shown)


jorenham:
cc @NeilGirdhar

image

NeilGirdhar:
Wow!!  Awesome work!!  You really are on the forefront of Python typing!

jorenham:
> Wow!! Awesome work!! You really are on the forefront of Python typing!

Thanks, that made my day :heart:

PR #1714, opened 2026-06-11
🐛`stats`: fix kappa type and rvs size in `vonmises_fisher` and `multivariate_normal` stubs

### LINKED ISSUE #1703: Various minor errors
Hi! I noticed two small typing issues in the SciPy stubs.

### 1. `vonmises_fisher_gen.__call__`: `kappa` should accept float-like inputs

Currently, `kappa` is annotated as `int` because of its default value, but it should accept floating point inputs:

```python
class vonmises_fisher_gen(multi_rv_generic):
    def __call__(
        self,
        /,
        mu: onp.ToFloat1D | None = None,
        kappa: onp.ToFloat1D = 1,
        seed: onp.random.ToRNG | None = None,
    ) -> vonmises_fisher_frozen: ...
```

This seems right conceptually, since `kappa` is a concentration parameter and should accept float-like scalar/array inputs.

### 2. `multivariate_normal_frozen.rvs`: `size` should accept empty tuple shape inputs

Currently:

```python
class multivariate_normal_frozen(
    multi_rv_frozen[multivariate_normal_gen],
    Generic[_ShapeT_co],
):
    @overload
    def rvs(
        self,
        /,
        size: onp.AtLeast0D,
        random_state: onp.random.ToRNG | None = None,
    ) -> onp.ArrayND[np.float64]: ...
```

For `rvs(size=...)`, `size` should accept empty tuple shape inputs as well as other tuple shapes. So this should be `onp.AtLeast0D` rather than an at-least-1D shape constraint.

Thanks!

Aniketsy:
fixes #1703


jorenham:
Looks good, thanks!

PR #1716, opened 2026-06-12
✨ `io.matlab._mio`: accept FileLike

myzhang1029:
This patch loosens the type annotations for the first argument of `mat_reader_factory`, `loadmat`, `savemat`, and `whosmat` to `FileLike[bytes]` from `FileName`.

- In internal uses of `mat_reader_factory` ([L241](https://github.com/scipy/scipy/blob/88776dae99516ccf07b66fdbb2f67091099ffc53/scipy/io/matlab/_mio.py#L241), [L395](https://github.com/scipy/scipy/blob/88776dae99516ccf07b66fdbb2f67091099ffc53/scipy/io/matlab/_mio.py#L395)), the first argument always gets an open file-like object rather than a path, so its signature probably should contain file-like objects.
-  The [docstring](https://github.com/scipy/scipy/blob/88776dae99516ccf07b66fdbb2f67091099ffc53/scipy/io/matlab/_mio.py#L287) and [code](https://github.com/scipy/scipy/blob/88776dae99516ccf07b66fdbb2f67091099ffc53/scipy/io/matlab/_mio.py#L325) of `savemat` both accept paths as well as file-like objects.

Remaining question:
- `loadmat`'s docstring does not mention that it accepts file-like objects, although the [code](https://github.com/scipy/scipy/blob/88776dae99516ccf07b66fdbb2f67091099ffc53/scipy/io/matlab/_mio.py#L240) does.
Should exclude `loadmat` from this patch? Or would it be fine for me to open another PR over `scipy` too? (I feel bad opening two consecutive DOC PRs 😅)

### DIFF
--- a/scipy-stubs/io/matlab/_mio.pyi
+++ b/scipy-stubs/io/matlab/_mio.pyi
@@ -5,7 +5,7 @@ from typing_extensions import deprecated
 import optype as op
 
 from ._miobase import MatFileReader
-from scipy.io._typing import ByteOrder, FileName
+from scipy.io._typing import ByteOrder, FileLike
 from scipy.sparse import coo_array, coo_matrix, csc_array, csc_matrix
 
 __all__ = ["loadmat", "savemat", "whosmat"]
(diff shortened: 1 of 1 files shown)


Aniketsy (commented):
thanks for the PR, looks fine to me, should we also add some test for this.

jorenham:
Thanks! This should be fine without additional tests; these annotations speak for themselves.

> * Or would it be fine for me to open another PR over `scipy` too? (I feel bad opening two consecutive DOC PRs 😅)

That's probably fine; consistent and complete documentation seems like something they'll appreciate, even if it requires two PRs to get there ;)

PR #1731, opened 2026-06-18
🐛✨ `signal`: `fft_mode`-dependent `ShortTimeFFT.istft` sctype

### LINKED ISSUE #1728: `scipy.signal.ShortTimeFFT.istft` return type annotation has a complex type but not a "real" type
[scipy.signal.ShortTimeFFT.istft](https://github.com/scipy/scipy/blob/52ee8ce715ed734f212ea8ef9a240feccf90d8fe/scipy/signal/_short_time_fft.py#L1457-L1479)

> It returns an array of dimension S.ndim - 1 which is real if onesided_fft is set, else complex. 

Pylance only shows a return type of `complex128`, which seems consistent with the stub file. 

https://github.com/scipy/scipy-stubs/blob/e77710fb55ec0b3a086299d13470670e6e47e089/scipy-stubs/signal/_short_time_fft.pyi#L352-L356

Can we add an overload or conditional definition to return float64 when `scipy.signal.ShortTimeFFT.fft_mode == 'onesided'`?

Thanks!

jorenham:
The `fft_mode` of `scipy.signal.ShortTimeFFT` affects the returned scalar type of its `istft` method. For one-sided modes it's `float64` and for two-sided ones it's `complex128`. This adds a new 2nd generic type parameter to `ShortTimeFFT` to capture this instance-scoped difference in `istft` return types. 

For the sake of type-safety this also changes the `ShortTimeFFT.setter` to no longer accept any values, effectively disallowing mutations, as this would violate the fundamental "types do not change" rule. At runtime this is still possible, but doing so would be type-unsafe (i.e. a bad idea).

Closes #1728

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,7 @@ See the `scipy` columns below for which classes are subscriptable at runtime.
 | generic type                                         | `scipy-stubs` | `scipy`  |                                                                                                 |
 | ---------------------------------------------------- | ------------- | -------- | ----------------------------------------------------------------------------------------------- |
 | `ShortTimeFFT[T: inexact]`                           | `>=1.16.0.0`  | `>=1.17` | [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.html)     |
+| `ShortTimeFFT[T: inexact, I: f64 \| c128]`           | `>=1.18.0.0`  | `>=1.17` | [docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.html)     |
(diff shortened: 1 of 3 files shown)


jorenham:
CC @hunterhogan; this was quite a bit more involved than I initially thought it would be haha. It seems to work as it should now though :)

I'll leave this open for a bit in case you (or anyone else for that matter) feels like taking a look.

hunterhogan:
I didn't know whether it would be difficult for an expert, but I could see that I had no idea how to condition a method's return's `dtype` on the class's initialization! Yikes!

I read the PR, so next time, I'll know that the way I can fix it is to open an Issue.

jorenham:
> I'll know that the way I can fix it is to open an Issue.

Much appreciated!

PR #1734, opened 2026-06-19
✅ `sparse`: add test coverage for `_bsr.pyi`

Aniketsy:
towards #1099


Aniketsy:
@jorenham could you please, ask once copilot for review in this .

Aniketsy:
Also, what is the fastest way to run these checks locally (i mean to ask how you check 😁 ), as if i use tox...(it get stucks for long time 5- 10 min sometimes) so i do test separately like `uv run pyrefly check tests/sparse/test_bsr.pyi`

jorenham:
> uv run pyrefly check

Hmm yea, tox is indeed meant for this, but it seems to be a bit broken right now. Lemme fix it real quick.

jorenham:
> Also, what is the fastest way to run these checks locally (i mean to ask how you check 😁 ), as if i use tox...(it get stucks for long time 5- 10 min sometimes) so i do test separately like `uv run pyrefly check tests/sparse/test_bsr.pyi`

Do you have lefthook enabled?

jorenham:
> > uv run pyrefly check
> 
> Hmm yea, tox is indeed meant for this, but it seems to be a bit broken right now. Lemme fix it real quick.

Ok if you rebase then `[uvx ]tox p` (`p` for parallel to speed things up) should now work again

Aniketsy:
> > Also, what is the fastest way to run these checks locally (i mean to ask how you check 😁 ), as if i use tox...(it get stucks for long time 5- 10 min sometimes) so i do test separately like `uv run pyrefly check tests/sparse/test_bsr.pyi`
> 
> Do you have lefthook enabled?

No, i tried to but i think something broken, i'll fix this 👍

Aniketsy:
> > 
> > Hmm yea, tox is indeed meant for this, but it seems to be a bit broken right now. Lemme fix it real quick.
> 
> Ok if you rebase then `[uvx ]tox p` (`p` for parallel to speed things up) should now work again

thanks ❤️ for quick fix (light speed 😸 )

Aniketsy:
```
tests/sparse/test_bsr.pyi:74: error: Expression is of type "bsr_array[numpy.bool[Any]]", not "bsr_array[numpy.bool[builtins.bool]]"  [assert-type]
tests/sparse/test_bsr.pyi:81: error: Expression is of type "bsr_matrix[numpy.bool[Any]]", not "bsr_matrix[numpy.bool[builtins.bool]]"  [assert-type]
```

could you please help with this... Do we need here `type ignore comment` ?

i've added previously when i run `just check`  and there was no error, but ci was pointing unused ignore comment so i've removed that and then again failure 😿

jorenham:
> ```
> tests/sparse/test_bsr.pyi:74: error: Expression is of type "bsr_array[numpy.bool[Any]]", not "bsr_array[numpy.bool[builtins.bool]]"  [assert-type]
> tests/sparse/test_bsr.pyi:81: error: Expression is of type "bsr_matrix[numpy.bool[Any]]", not "bsr_matrix[numpy.bool[builtins.bool]]"  [assert-type]
> ```
> 
> 
>     
> 
>   
> 
> could you please help with this... Do we need here `type ignore comment` ?

Ah... that looks like the same regression in mypy 2.2.0 as the one causing issues in https://github.com/numpy/numpy/pull/32024. Here we can't just listen to mypy though, because the other type-checkers will disagree (as they should; it's a regression, i.e. a bug, after all).

I think the cleanest workaround here would be to replace these failing `assert_type` with `optype.test.assert_subtype` (if you're not yet familiar, ctrl+F the tests to see how to use it).

tests/sparse/test_bsr.pyi
```diff
@@ -0,0 +1,180 @@
...
+assert_type(bsr_array(_seq_seq_int, blocksize=(2, 2)), bsr_array[np.int_])  # type: ignore[assert-type]
+assert_type(bsr_array(_data2, blocksize=(2, 2)), bsr_array[ScalarType])
+assert_type(bsr_matrix(_seq_seq_int, blocksize=(2, 2)), bsr_matrix[np.int_])  # type: ignore[assert-type]
+assert_type(bsr_matrix(_data2, blocksize=(2, 2)), bsr_matrix[ScalarType])
+
+###
+# (M, N) shape constructor, dtype positional (requires shape positional too)
+
+assert_type(bsr_array(_shape2, _shape2, np.float64), bsr_array[np.float64])
+assert_type(bsr_array(_shape2, _shape2, np.bool_), bsr_array[np.bool_])
```
jorenham: pedantic nit:

```suggestion
assert_type(bsr_array(_shape2, _shape2, np.bool), bsr_array[np.bool])
```

I kinda wish there was a ruff rule for this 🤔

jorenham: There's some others like this above as well, I see

jorenham:
overall this looks great, btw

tests/sparse/test_bsr.pyi
Aniketsy: > I kinda wish there was a ruff rule for this 🤔

will it make sense to create this 🤷 (i've stolen this emozi from you sorry )

jorenham:
thanks!

PR #1949, opened 2026-07-10
🐛 `datasets`: restrict `face(gray=1)` to 3-D overload to match `gray is True`

### LINKED ISSUE #1842: `datasets`: `face(gray=1)` output ndim depends on `1` vs `True`
Claude Fable 5 reports:

- [`datasets/_fetchers.pyi:17`](scipy-stubs/datasets/_fetchers.pyi#L17) `face(gray=1)` typed as the 2-D overload via `L[True, 1]`, but the implementation checks `gray is True`, so `1` gives the 3-D color image (`0` is fine → `L[True]` / `L[False, 0]`).

Aniketsy:
fixes #1842


scipy-stubs/datasets/_fetchers.pyi
```diff
@@ -14,6 +14,6 @@ def fetch_data(
 def ascent() -> AscentDataset: ...
 def electrocardiogram() -> ECGDataset: ...
 @overload
-def face(gray: L[True, 1]) -> Face2Dataset: ...
+def face(gray: L[True]) -> Face2Dataset: ...  # type: ignore[overload-overlap]
```
jorenham: let's just use `L[False]` below so we don't have to ignore this mypy error

jorenham: (unless there's an actual usecase for 1 and 0)

Aniketsy: > let's just use L[False] below so we don't have to ignore this mypy error

hmm make sense i did not find usecase

jorenham:
thanks!

PR #1964, opened 2026-07-11
🐛 `ndimage`: fix `labeled_comprehension` return type for index=None/scalar

### LINKED ISSUE #1828: `ndimage`: `labeled_comprehension(index=None)` output should be scalar,  not Nd
Claude Fable 5 reports:

- **`labeled_comprehension(index=None)`** ([`ndimage/_measurements.pyi:71-110`](scipy-stubs/ndimage/_measurements.pyi#L71)): claimed `ArrayND`; runtime returns the bare `func` result (scalar).

Aniketsy:
Fixes #1828

### DIFF
--- a/scipy-stubs/ndimage/_measurements.pyi
+++ b/scipy-stubs/ndimage/_measurements.pyi
@@ -72,7 +72,17 @@ def value_indices(
 def labeled_comprehension[ScalarT: npc.number | np.bool](
     input: onp.ToComplex | onp.ToComplexND,
     labels: onp.ToComplex | onp.ToComplexND | None,
-    index: onp.ToInt | onp.ToIntND | None,
+    index: onp.ToInt | None,
+    func: _ComprehensionFunc,
+    out_dtype: _DTypeLike[ScalarT],
+    default: onp.ToFloat,
+    pass_positions: bool = False,
+) -> ScalarT: ...
+@overload
+def labeled_comprehension[ScalarT: npc.number | np.bool](
+    input: onp.ToComplex | onp.ToComplexND,
+    labels: onp.ToComplex | onp.ToComplexND | None,
+    index: onp.ToIntND,
     func: _ComprehensionFunc,
     out_dtype: _DTypeLike[ScalarT],
     default: onp.ToFloat,
@@ -82,7 +92,17 @@ def labeled_comprehension[ScalarT: npc.number | np.bool](
 def labeled_comprehension(
     input: onp.ToComplex | onp.ToComplexND,
     labels: onp.ToComplex | onp.ToComplexND | None,
-    index: onp.ToInt | onp.ToIntND | None,
+    index: onp.ToInt | None,
+    func: _ComprehensionFunc,
+    out_dtype: onp.AnyIntPDType,
+    default: onp.ToInt,
+    pass_positions: bool = False,
+) -> np.intp: ...
+@overload
+def labeled_comprehension(
+    input: onp.ToComplex | onp.ToComplexND,
+    labels: onp.ToComplex | onp.ToComplexND | None,
+    index: onp.ToIntND,
     func: _ComprehensionFunc,
     out_dtype: onp.AnyIntPDType,
     default: onp.ToInt,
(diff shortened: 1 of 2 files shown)


Aniketsy:
I've splitted the overload for `scalar` and `array` output . please let me know if you will prefer different approach for this.

scipy-stubs/ndimage/_measurements.pyi
```diff
@@ -72,7 +72,17 @@ def value_indices(
 def labeled_comprehension[ScalarT: npc.number | np.bool](
     input: onp.ToComplex | onp.ToComplexND,
     labels: onp.ToComplex | onp.ToComplexND | None,
-    index: onp.ToInt | onp.ToIntND | None,
+    index: onp.ToInt | None,
+    func: _ComprehensionFunc,
+    out_dtype: _DTypeLike[ScalarT],
```
jorenham, line 77: Hmm, it turns out that when `index=None`, the `out_dtype` is completely ignored (only for `None`;  int scalar index respects `out_dtype`). So although this issue was pre-existing, we might as well fix this as well here by splitting this overload into two.

tests/ndimage/test__measurements.pyi
```diff
@@ -68,15 +68,29 @@ assert_type(value_indices(int_2d), dict[np.intp, tuple[onp.ArrayND[np.intp], ...
...
+
+# out_dtype: DTypeLike[_SCT] -> _SCT
+assert_type(labeled_comprehension(f64_2d, None, None, _stat_func, np.dtype(np.float32), 0.0), np.float32)
+# out_dtype: AnyIntPDType -> np.intp
+assert_type(labeled_comprehension(f64_2d, None, None, _stat_func, np.intp, 0), np.intp)
+# out_dtype: AnyFloat64DType | None -> np.float64
+assert_type(labeled_comprehension(f64_2d, None, None, _stat_func, None, 0.0), np.float64)
+assert_type(labeled_comprehension(f64_2d, None, None, _stat_func, np.float64, 0.0), np.float64)
+# out_dtype: AnyComplex128DType -> np.complex128
+assert_type(labeled_comprehension(f64_2d, None, None, _stat_func, np.complex128, 0.0), np.complex128)
```
jorenham, line 82: As I mentioned above, these assertions don't match runtime behavior now.

tests/ndimage/test__measurements.pyi
```diff
@@ -68,15 +68,29 @@ assert_type(value_indices(int_2d), dict[np.intp, tuple[onp.ArrayND[np.intp], ...
...
+
+# index=<int array> -> ArrayND output
```
jorenham, line 84: we should then also test the scalar `index` case separately here

scipy-stubs/ndimage/_measurements.pyi
```diff
@@ -72,7 +72,17 @@ def value_indices(
...
+    default: onp.ToFloat,
+    pass_positions: bool = False,
+) -> ScalarT: ...
+@overload
+def labeled_comprehension[ScalarT: npc.number | np.bool](
+    input: onp.ToComplex | onp.ToComplexND,
+    labels: onp.ToComplex | onp.ToComplexND | None,
```
jorenham, line 84: Another pre-existing issue here is that when `index is not None` and `labels is None`, a `ValueError` is raised at runtime. So it might be nice to prevent that by removing `| None` for the not-`None` index overloads like this one.

jorenham (commented):
The overload split itself is indeed the way to go here.

There seem to be some (other) pre-existing issues here that can now also (easily) be fixed now that these index cases are split up. We could do that here or as a follow-up; I have no particular preference.

Aniketsy:
> The overload split itself is indeed the way to go here.
> 
> There seem to be some (other) pre-existing issues here that can now also (easily) be fixed now that these index cases are split up. We could do that here or as a follow-up; I have no particular preference.

if this is clean, how about fixing in follow up

jorenham:
> > The overload split itself is indeed the way to go here.
> > There seem to be some (other) pre-existing issues here that can now also (easily) be fixed now that these index cases are split up. We could do that here or as a follow-up; I have no particular preference.
> 
> if this is clean, how about fixing in follow up

The assertions for `index=None` you added are mostly incorrect, so if you could add a TODO comment stating that it'll be fixed in a followup, then this is good to merge

tests/ndimage/test__measurements.pyi
```diff
@@ -68,15 +68,29 @@ assert_type(value_indices(int_2d), dict[np.intp, tuple[onp.ArrayND[np.intp], ...
 ###
 # labeled_comprehension
 
+# index=None -> scalar output
+
```
Aniketsy:
```suggestion
# index=None -> scalar output
# TODO: assertions below are mostly incorrect — out_dtype is ignored at runtime when index=None.
# Also missing a separate test for scalar index (e.g. index=1). Will fix in follow-up.
```

Aniketsy: will this be fine ? i did not tag you in TODO, as i'll address these

jorenham: Yea seems fine

PR #1988, opened 2026-07-15
:bug: `stats`: Propagate `quantile` floating dtypes

### LINKED ISSUE #1897: `stats`: `quantile` should propagate floating dtypes
Claude Fable 5 reports:

[...] Dtype is also input-tracking (`quantile(f32, .5)` → `np.float32`), not always float64.

related: #1776

sapunyangkut:
## Summary

- preserve `np.float32` return dtypes when `x`, `p`, and optional weights are float32-compatible
- retain float64 promotion when `p` or weights require it
- cover scalar/array probabilities, axis handling, `keepdims`, and weighted quantiles with `assert_type` tests

## Root cause

The existing `quantile` overloads always fell through to float64 return annotations, even though SciPy derives the runtime dtype from the floating inputs. Ordered float32 overloads now model that behavior while the existing float64 overloads remain the promotion fallback.

Fixes #1897

## Validation

- full repository `just check` passed: ruff, dprint, typos, zizmor, ty, zuban, pyrefly, mypy, basedpyright, and stubtest (1010 modules)
- runtime probe passed on Python 3.12.10, NumPy 2.5.1, and SciPy 1.18.0 for scalar/array probabilities, `axis`, `keepdims`, and float16/float32/float64 weights
- `git diff --check`
- changed-diff secret scan passed; the static stub and `assert_type` test diff has no Rafter security surface or dependency change

## AI assistance

This change was implemented and validated with OpenAI Codex assistance. The validation details above are machine-generated and disclosed for maintainer review.


jorenham:
Could you, as a human, walk me through these changes without relying on AI tools? I just want to make sure that you understand it yourself, you see :)

jorenham:
closing in favor of #2191

PR #1993, opened 2026-07-16
🐛 `ndimage`: `labeled_comprehension` ignores `out_dtype` when `index=None`; reject `labels=None` with non-`None` `index`

Aniketsy:
follow up #1964 

> Hmm, it turns out that when index=None, the out_dtype is completely ignored (only for None; int scalar index respects out_dtype). So although this issue was pre-existing, we might as well fix this as well here by splitting this overload into two.

>Another pre-existing issue here is that when index is not None and labels is None, a ValueError is raised at runtime. So it might be nice to prevent that by removing | None for the not-None index overloads like this one.

### DIFF
--- a/tests/ndimage/test__measurements.pyi
+++ b/tests/ndimage/test__measurements.pyi
@@ -43,6 +43,7 @@ out_labeled: onp.ArrayND[np.int32 | np.intp]
 
 # helper function for labeled_comprehension
 def _stat_func(x: onp.ToComplex | onp.ToComplexND) -> np.float64: ...
+def _stat_func_with_positions(x: onp.ToComplex | onp.ToComplexND, positions: onp.ToComplex | onp.ToComplexND) -> np.float64: ...
 
 ###
 # label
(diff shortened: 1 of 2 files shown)


tests/ndimage/test__measurements.pyi
jorenham, line 1: it might be worth adding a testcase for the fancy new `_ComprehensionFunc` overload as well

tests/ndimage/test__measurements.pyi
```diff
@@ -69,18 +69,24 @@ assert_type(value_indices(int_2d), dict[np.intp, tuple[onp.ArrayND[np.intp], ...
 ###
 # labeled_comprehension
 
-# index=None -> scalar output
-# TODO: assertions below are mostly incorrect — out_dtype is ignored at runtime when index=None.
-# Also missing a separate test for scalar index (e.g. index=1). Will fix in follow-up.
+# index=None -> the return type of `func`, as-is; out_dtype and default are ignored at runtime
```
jorenham: pedantic nit: backtick ` use is inconsistent in this comment

jorenham (approved):
I left two optional suggestions, but I'm also fine with merging this as-is if you want

Aniketsy:
thanks for the review ..

> I left two optional suggestions, but I'm also fine with merging this as-is if you want

sure, lets move this toward perfection . i'll update with testcase and inconsistent backtick ` use in the comment

then we can merge this .

jorenham:
thanks!

Issue #1940, opened 2026-07-09
`linalg`: improve return dtypes of `pinv[h]`, `matrix_balance`, and `matmul_toeplitz`


jorenham:
See the TODO's in `scipy-stubs/linalg/_basic.pyi `.

- [x] `pinv` and `pinvh`
- [x] `matrix_balance` 
- [x] `matmul_toeplitz`

Aniketsy:
> TODO(@jorenham): better naming

please let me know if this sounds good, or if you're planning to do by yourself i'm fine too, if this needs more improvements.

- _InputFloat  -> _ToFloat
- _InputFloatStrict1D -> _ToFloatStrict1D
- _InputFloatStrict2D -> _ToFloatStrict2D
- _InputF64 -> _ToFloat64
- _InputF64Strict1D -> _ToFloat64Strict1D
- _InputF64Strict2D -> _ToFloat64Strict2D
- _InputComplex -> _ToComplex
- _InputComplexStrict1D -> _ToComplexStrict1D
- _InputComplexStrict2D   -> _ToComplexStrict2D

jorenham:
`To*` types shouldn't be used for output, so your proposed overloads are too broad. And except for `float64`, the other scalar supported types are also missing.

Aniketsy:
> `To*` types shouldn't be used for output, so your proposed overloads are too broad. And except for `float64`, the other scalar supported types are also missing.

oh okay then, its better to leave this for you

PR #1997, opened 2026-07-18
✨ `linalg.pinv[h]`: improved shape-typing and dtype propagation

jorenham:
The `scipy.linalg.pinv` and `pinvh` functions are now overloaded to exactly cover each 4 (`f32`/`f64`/`c64`/`c128`) output dtypes, as well as 2d and 3d shape-types (including fallbacks for >3d of course).

towards #1940


jorenham:
cc @Aniketsy

PR #2005, opened 2026-07-19
🦖 `stats`: `rvs` fallback overloads for compatibility with legacy type-checkers

### LINKED ISSUE #1702: Add fallback overloads for indeterminate length size
Hi! I think the current `rvs` overloads are length-precise, but they reject `size: tuple[int, ...]` even though their domains collectively cover all tuple lengths.

As far as I understand, this is expected checker behavior: overload resolution does not generally prove coverage like:

```python
tuple[int, ...] <= (
    tuple[()]
    | tuple[int]
    | tuple[int, int]
    | tuple[int, int, int, *tuple[int, ...]]
)
```

It instead requires compatibility with an individual overload. Since `tuple[int, ...]` spans multiple overloads, none of the precise overloads is applicable.

Would it make sense to add broad fallback overloads after the precise ones? That should preserve precise shape inference when available, while still accepting ordinary erased runtime shapes.

For `dirichlet_frozen.rvs`:

```python
@overload
def rvs(
    self,
    /,
    size: tuple[SupportsIndex, ...],
    random_state: onp.random.ToRNG | None = None,
) -> onp.ArrayND[np.float64]: ...
```

For `multivariate_normal_frozen.rvs`:

```python
@overload
def rvs(
    self,
    /,
    size: tuple[int, ...],
    random_state: onp.random.ToRNG | None = None,
) -> onp.ArrayND[np.float64]: ...
```

Thanks!

jorenham:
Turns out my worries about overlapping overloads leading leading to `Any` results were unfounded; adding fallback overloads for `size` doesn't seem to cause problems :tada:.

Closes #1702

### DIFF
--- a/scipy-stubs/stats/_multivariate.pyi
+++ b/scipy-stubs/stats/_multivariate.pyi
@@ -980,6 +980,10 @@ class dirichlet_gen(multi_rv_generic):
         size: tuple[SupportsIndex, SupportsIndex, SupportsIndex, *tuple[SupportsIndex, ...]],
         random_state: onp.random.ToRNG | None = None,
     ) -> _Array3ND[np.float64]: ...
+    @overload
+    def rvs(
+        self, /, alpha: onp.ToFloat1D, size: tuple[SupportsIndex, ...], random_state: onp.random.ToRNG | None = None
+    ) -> _Array1ND[np.float64]: ...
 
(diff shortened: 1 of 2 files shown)


jorenham:
cc @NeilGirdhar

PR #2006, opened 2026-07-20
✅ `stats.qmc`: add test coverage for `Halton`

Aniketsy:
Towards #1099


tests/stats/test__qmc.pyi
```diff
@@ -0,0 +1,30 @@
...
+
+_h4 = Halton(_d, rng=0)
+assert_type(_h4, Halton)
+
+_h5 = Halton(_d, seed=0)
+assert_type(_h5, Halton)
+
+assert_type(_h1.base, list[int])
+assert_type(_h1.scramble, bool)
+assert_type(_h1._permutations, list[onp.Array2D[np.int_]])  # noqa: SLF001
```
Aniketsy: not sure if you want to add test for this as its private member, i've suppressed the warning for now, please let me know if you want me to remove this , :)

tests/stats/test__qmc.pyi
```diff
@@ -0,0 +1,30 @@
...
+assert_type(_h2, Halton)
+
+_h3 = Halton(_d, scramble=False)
+assert_type(_h3, Halton)
+
+_h4 = Halton(_d, rng=0)
+assert_type(_h4, Halton)
+
+_h5 = Halton(_d, seed=0)
+assert_type(_h5, Halton)
```
jorenham: these intermediate `_h*` are only used once, so it would be cleaner to inline them

tests/stats/test__qmc.pyi
jorenham, line 1: `stats.qmc` is a subpackage, and the tests live in a corresponding subdirectory `/tests/stats/qmc` (https://github.com/scipy/scipy-stubs/tree/31f85074109b26e4fcd0a84308d08a2c683de290/tests/stats/qmc)

Aniketsy: ahh, i did search 😿 , sorry i missed 

lol, i only added that file 😆

tests/stats/test__qmc.pyi
Aniketsy: make sense 👍

tests/stats/test__qmc.pyi
jorenham: I don't mind, but there's no real need to either

Aniketsy: hmm, you left me in between 😄 whether i should say, lets keep it or lets remove it .... fine for the sense of completeness

tests/stats/test_qmc.pyi
jorenham, line 1: the `test_scale` module name isn't really accurate now. So we could add this as a separate `test_halton.pyi` or rename this module.

Aniketsy: if we go with renaming , what would you suggest as this includes two tests, and if needs to add in separate `test_halton.pyi` . i'll update shortly...i'm in mid of other changes

jorenham: if we'd rename then I think we should flatten this to `tests/stats/test_qmc.pyi`

Aniketsy: lets go with this then 👍

PR #2049, opened 2026-07-26
🔥 remove copilot instructions

jorenham:
It wasn't really very helpful either way, and I'm not using copilot anymore myself.


Aniketsy:
> It wasn't really very helpful either way, and I'm not using copilot anymore myself.

make sense to remove this 👍

PR #2083, opened 2026-07-31
✅`stats`: add test coverage for `chisquare`

Aniketsy:
Towards #1099

### DIFF
--- a/tests/stats/test__stats_py.pyi
+++ b/tests/stats/test__stats_py.pyi
@@ -1061,8 +1061,24 @@ assert_type(power_divergence(_f64_nd, keepdims=True), Power_divergenceResult[onp
 # chisquare
 
 assert_type(chisquare(_f64_1d), Power_divergenceResult[np.float64])
-assert_type(chisquare(_f64_nd, axis=None), Power_divergenceResult[np.float64])
-assert_type(chisquare(_f64_nd, keepdims=True), Power_divergenceResult[onp.ArrayND[np.float64]])
+assert_type(chisquare(_f64_1d, _f64_1d), Power_divergenceResult[np.float64])
+assert_type(chisquare(_f64_1d, ddof=1), Power_divergenceResult[np.float64])
+assert_type(chisquare(_f64_1d, axis=None), Power_divergenceResult[np.float64])
+assert_type(chisquare(_f64_1d, axis=1), Power_divergenceResult[np.float64])
+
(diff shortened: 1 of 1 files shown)


Aniketsy:
>@deprecated("will be removed in SciPy v2.0.0")
def compare_medians_ms(group_1: object, group_2: object, axis: object = None) -> object: ...

Does it make sense to add test coverage for deprecated items, which will be removed eventually.

I don't think so

jorenham:
that looks like the `compare_medians_ms` from the `mstats_extras` module, which indeed will be gone next release, and is therefore excluded from the test coverage totals:

https://github.com/scipy/scipy-stubs/blob/d7937b8f8d97cb9cfde66592e84b6be79a0f84d0/scripts/test_coverage.py#L169

tests/stats/test__stats_py.pyi
```diff
@@ -1124,23 +1061,28 @@ assert_type(power_divergence(_f64_nd, keepdims=True), Power_divergenceResult[onp
...
+assert_type(chisquare(_f64_2d, f_exp=_f64_2d, ddof=1, axis=None), Power_divergenceResult[np.float64])
+
+assert_type(chisquare(_f64_2d, keepdims=True), Power_divergenceResult[onp.ArrayND[np.float64]])
+assert_type(chisquare(_f64_2d, axis=1, keepdims=True), Power_divergenceResult[onp.ArrayND[np.float64]])
+assert_type(chisquare(_f64_2d, axis=None, keepdims=True), Power_divergenceResult[onp.ArrayND[np.float64]])
+
+assert_type(chisquare(_f64_2d), Power_divergenceResult[np.float64 | Any])
+assert_type(chisquare(_f64_2d, axis=0), Power_divergenceResult[np.float64 | Any])
+assert_type(chisquare(_f64_2d, axis=1), Power_divergenceResult[np.float64 | Any])
+assert_type(chisquare(_f64_2d, keepdims=False), Power_divergenceResult[np.float64 | Any])
```
jorenham, line 1081: these test `stats.chisquare`, not `stats.mstats.chisquare`

Aniketsy: https://github.com/scipy/scipy-stubs/blob/e837293c819d8f4d1fe7300d12e20809b5ab153f/scipy-stubs/stats/mstats.pyi#L3

we don't have overloads for `stats.mstats.chisquare`

jorenham: Ah I see. We should  exclude those in the test coverage script then.

jorenham:
thanks

PR #2088, opened 2026-08-03
🔨 exclude `mstats` re-exports of `scipy.stats` functions from test coverage

jorenham:
Originally noted by @Aniketsy in https://github.com/scipy/scipy-stubs/pull/2083#discussion_r3701699783


Aniketsy:
Thanks for the fix :)

PR #2331, opened 2026-08-31
experiment: expand dtype unions in return type annotations.

randolf-scholz:
[check_return_arrays.py](https://github.com/user-attachments/files/31658248/check_return_arrays.py)

The vibecoded script above was applied with `python check_return_arrays.py scipy-stubs/ --fix`.

its purpose is to expand return type annotations like ` -> Array[X | Y]` to `-> Array[X] | Array[Y]`.

Tests were not patched, this is just to check mypy primer output.


jorenham:
No mypy_primer diff :tada:

randolf-scholz:
That should be somewhat expected since this was a very conservative change, and in many places return types that are arrays of dtype unions are hidden behind aliases. The danger with modifying those aliases however is that they may also be used in argument positions, and then the expansion would be a breaking change.

jorenham:
> The danger with modifying those aliases however is that they may also be used in argument positions, and then the expansion would be a breaking change.

We should indeed avoid those, and limit this to covariant/output positions only

jorenham:
Were you planning on working some more on this? Otherwise I could take over If you'd like :)

PR #2371, opened 2026-09-11
⬆️ pyrefly 1.3.0

jorenham:
https://github.com/facebook/pyrefly/releases/tag/1.3.0


jorenham:
@rchen152, could these new failures be related to the recent overload changes?

rchen152:
Yeah, the `assert_type(Unknown, ...)` errors look like the overload change. https://github.com/scipy/scipy-stubs/actions/runs/34589419016/job/103231022161?pr=2371 also has some inconsistent override errors, which are due to us now checking for inconsistent `__call__` method overrides when inheriting from a Protocol. (Previously we ignored `__call__` altogether for inconsistent override checks.)

